PR #2, opened 2017-08-27
fix: clean up optional parameters signature

joscha:
Before, it was possible to pass an undefined `storyFn`.


timhaines:
Thanks @joscha - published as version 1.1.5.

PR #12, opened 2017-10-05
Fix when script tag includes type attribute

Peeja:
Not clear to me why my Storybook install is emitting `type="text/javascript` as part of the `script` tag, as I don't see code that does it in Storybook or a change that introduced it. But it doesn't seem like it hurts to accept it.

Without this, the `percy-storybook` command fails with:

```
Error:  TypeError: Cannot read property '1' of null
```


timhaines:
Thanks for this PR @Peeja!  We tweaked the regex a little, added some tests, and shipped it as v1.2.4 

https://github.com/percy/percy-storybook/releases/tag/%40percy-io%2Fpercy-storybook%401.2.4

Peeja:
Cheers!

Issue #20, opened 2017-12-13
Fix stripping of ./ from the start of build_dir for static assets when included


timhaines:
When we prepare static assets, we strip the build_dir from the path, [here](https://github.com/percy/percy-storybook/blob/master/packages/percy-storybook/src/getStaticAssets.js#L36).

This fails to work correctly when the build_dir is prefixed with ./

timhaines:
Fixed as part of https://github.com/percy/percy-storybook/pull/21

PR #19, opened 2017-12-12
Allow JSON output from percy-storybook

joscha:
This prints something like:
```json
{
  "branch": "master",
  "build-number": 327,
  "web-url": "https://percy.io/XXX/test/builds/123",
  "state": "pending",
  "is-pull-request": false,
  "pull-request-number": null,
  "pull-request-title": null,
  "user-agent": "Percy/v1 percy-storybook/1.2.7 percy-js/2.4.2 (storybook/unknown react/unknown; node/v8.7.0)",
  "total-comparisons-finished": null,
  "total-comparisons-diff": null,
  "failure-reason": null,
  "parallel-nonce": null,
  "parallel-total-shards": null,
  "finished-at": null,
  "approved-at": null,
  "created-at": "2017-12-12T05:40:44.000Z",
  "updated-at": "2017-12-12T05:40:44.000Z"
}
```

PR #21, opened 2017-12-13
Add output_format arg, that provides for JSON output of build status

timhaines:
Thanks @joscha for doing the majority of the work here in https://github.com/percy/percy-storybook/pull/19.

Also includes a fix for https://github.com/percy/percy-storybook/issues/20.

### DIFF
--- a/packages/percy-storybook/src/cli.js
+++ b/packages/percy-storybook/src/cli.js
@@ -51,7 +51,9 @@ export async function run(argv) {
 
   if (process.env.PERCY_ENABLE === '0') {
     // eslint-disable-next-line no-console
-    console.log(options.json ? '{}' : 'The PERCY_ENABLE environment variable is set to 0. Exiting.');
+    console.log(
+      options.json ? '{}' : 'The PERCY_ENABLE environment variable is set to 0. Exiting.',
+    );
     return;
   }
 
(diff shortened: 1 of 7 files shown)


joscha:
What do you think about making the format parameter lowercase or accepting it case-independent.

timhaines:
@joscha Done.  output_format now accepts JSON or json.

packages/percy-storybook/src/cli.js
```diff
@@ -66,13 +74,19 @@ export async function run(argv) {
   // debug('assets %o', assets);
 
   const stories = await getStories(assets[storybookJavascriptPath], options);
-  debug('stories %o', stories);
+  options.debug && debug('stories %o', stories);
```
fotinakis: This toggling of output should be the job of the `debug` function itself, I believe you can just do `debug.enabled = options.debug` to toggle it on or off. https://www.npmjs.com/package/debug#checking-whether-a-debug-target-is-enabled

packages/percy-storybook/src/getOutputFormat.js
```diff
@@ -0,0 +1,11 @@
+export default function getOutputFormat(outputFormatString) {
```
fotinakis: Maybe `normalizeOutputFormat`?

PR #22, opened 2017-12-28
Fix rtl regex

timhaines:
We had a bug where using the g flag in the rtl regex was making the regex sticky, and causing matches on subsequent names that were shorter to fail.  

The result was some RTL stories were being skipped when the --rtl option was being used.


joscha:
cc @marcfallows this will fix the strange bug that we encountered.

PR #25, opened 2018-01-05
Update integration tests to Storybook 3.3.3

timhaines:
Previously used Storybook 3.2.x.


timhaines:
Percy snapshots show some failures here.  I need to investigate what's caused them.

timhaines:
Storybook 3.3 seems to have broken the feature that imports preview-head.html into iframe.html.  Issue opened: https://github.com/storybooks/storybook/issues/2665

PR #36, opened 2018-05-13
 Handle Storybook 4.0 code-splitting preview.js into multiple js files

timhaines:
Storybook previously created one javascript file, static/preview-xxx.js, which could be executed to retrieve all of the story names.

Storybook 4.0 introduces code-splitting, so multiple static/*.js files need to be executed in order to retrieve all of the story names.

I've updated the React integration tests to use Storybook 4.0, and have left the Vue and Angular ones at 3.4.

### DIFF
--- a/integration-tests/storybook-for-react/stories/index.js
+++ b/integration-tests/storybook-for-react/stories/index.js
@@ -1,6 +1,8 @@
 import React from 'react';
-import { storiesOf, action } from '@storybook/react';
+import { storiesOf } from '@storybook/react';
 import { withInfo } from '@storybook/addon-info';
+import { action } from '@storybook/addon-actions';
+import { Button } from '@storybook/react/demo';
 import faker from 'faker';
 
 // Disabling the two lines below until react-match-media that works with react 16 is released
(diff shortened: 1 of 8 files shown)


integration-tests/storybook-for-react/stories/index.js
```diff
@@ -1,15 +1,23 @@
...
 // import { MatchMediaHOC } from 'react-match-media';
 // import Example from '../src/Example';
 
 storiesOf('Button', module)
-  .add('with text', () => <button onClick={action('clicked')}>Hello Button</button>)
-  .add('with some emoji', () => <button onClick={action('clicked')}>😀 😎 👍 💯</button>);
+  .add('with text', () => <Button onClick={action('clicked')}>Hello Button</Button>)
+  .add('with some emoji', () => (
+    <Button onClick={action('clicked')}>
+      <span role="img" aria-label="so cool">
```
djones: I'm curious what this span adds to the test. Is it making use of or testing some Storybook 4.x features?

timhaines: @djones These two button stories are a copy of the default storybook stories.  Their new v4 storybook generates them like this, so I copy/pasted.  I don't have an understanding of it, but I'd guess it's a convenient way of getting 'so cool' in there for screen-readers (the aria label).

timhaines: @djones Yeah, I think that's it.  Here's the Storybook commit it was added in: https://github.com/storybooks/storybook/commit/e5b73a3d73d5b9251e8386ab3226d73564ecbc81

PR #39, opened 2018-05-26
Use puppeteer to read stories from static-storybook instead of jsdom

timhaines:
One job percy-storybook has to do is open and execute the storybook's javascript to get the list of stories.

Before this PR, percy-storybook used jsdom to open the built storybook's javascript files.  This came with several problems 
1) jsdom is a feature-incomplete browser implementation, so stories/javascript that use advanced browser features would break the javascript execution, and cause percy-storybook to fail.
2) some custom mock code was used to patch in missing browser features, but this is adhoc and like whack-a-mole -- an incomplete solution.
3) It's possible that some inline JS in the preview-head wasn't being executed, causing percy-storybook to fail.  This problem is also mitigated by this PR.

After this PR, jsdom is no longer used, and instead the entire static storybook is opened with puppeteer/Chrome, and the stories are read.  This solves each of the 3 problems above.

One change to note is that when running on linux, Chrome is opened with the --no-sandbox flag.  This is required on Travis, and possibly some other CI providers and in docker. 

Also refactors out the `getIframePath` method which is used in two locations.

I plan to ship this as a major version bump, as it's a significant change - and may possibly break some people's CI processes in unforeseen ways.

TODO:
- [x] Fix tests for Appveyor/Windows


packages/percy-storybook/src/getStories.js
```diff
@@ -1,111 +1,70 @@
...
-          store[key] = value.toString();
-        },
-        clear: function() {
-          store = {};
-        }
-      };
-    })();
-    Object.defineProperty(window, 'localStorage', { value: localStorageMock });
-    `;
+  // Travis CI requires Chrome to be launched without the sandbox
```
joscha: I this only true for Travis or other CI systems as well? In the latter, it would probably be good to check `process.env.CI` which a lot of different vendors set to `true`.

packages/percy-storybook/src/getStories.js
```diff
@@ -1,111 +1,70 @@
...
-const matchMediaMock = `
-    window.matchMedia = window.matchMedia || (() => {
-      return {
-        matches: false,
-        addListener: () => {},
-        removeListener: () => {},
-      };
-    });
-    `;
+  const browser = await puppeteer.launch({ headless: true, args: launchArgs });
```
joscha: This will now require the user to have some sort of framebuffer, right? E.g. before it wouldn't need anything, now it needs some sort of `DISPLAY=...` and possibly Xvfb or such.

timhaines: @joscha puppeteer will use Chromium (or Chrome) Headless by default, so no framebuffer is required.

packages/percy-storybook/src/getStories.js
timhaines: @joscha I'm going to update this PR to make --no-sandbox the default.

packages/percy-storybook/src/__tests__/getIframePath-tests.js
```diff
@@ -0,0 +1,28 @@
...
+});
+
+it('returns the iframe path when the storybook exists', () => {
+  mock({
+    '/tmp-test': {
+      'iframe.html': 'file content here',
+    },
+  });
+  try {
+    if (process.platform === 'win32') {
```
djones: Nice Windows handling here

packages/percy-storybook/src/__tests__/getStories-tests.js
```diff
@@ -1,65 +1,31 @@
 import getStories from '../getStories';
-import { storiesKey } from '../constants';
+// import { storiesKey } from '../constants';
```
djones: Safe to delete this now?

packages/percy-storybook/package.json
```diff
@@ -22,15 +22,16 @@
...
+    "babel-register": "^6.26.0",
+    "mock-fs": "^4.5.0"
   },
   "dependencies": {
     "@percy/react-percy-api-client": "^0.4.3",
     "babel-runtime": "^6.26.0",
     "debug": "^3.1.0",
     "es6-error": "^4.0.2",
     "es6-promise-pool": "^2.4.4",
-    "jsdom": "^9.12.0",
```
djones: 👍

packages/percy-storybook/src/getStories.js
```diff
@@ -1,111 +1,70 @@
...
-          store = {};
-        }
-      };
-    })();
-    Object.defineProperty(window, 'localStorage', { value: localStorageMock });
-    `;
+  // Some CI platforms including Travis requires Chrome to be launched without the sandbox
+  // See https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-travis-ci
+  // See https://docs.travis-ci.com/user/chrome#Sandboxing
+  if (os.platform() === 'linux') {
```
djones: :+1: for limiting this to just linux.

packages/percy-storybook/src/__tests__/getStories-tests.js
```diff
@@ -1,65 +1,31 @@
...
 
 let originalTimeout = 0;
 
 beforeEach(function() {
   originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL;
-  jasmine.DEFAULT_TIMEOUT_INTERVAL = 12000;
+  jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000;
```
djones: Curious... I assume you wanted to make this bigger because loading a headless browser will likely be a bit slower than `jsdom`?

packages/percy-storybook/src/__tests__/getStories-tests.js
```diff
@@ -1,65 +1,31 @@
...
 
   expect.assertions(1);
 });
 
 it('returns the value window[storiesKey] is set to', async () => {
-  const code = `if (typeof window === 'object') window['${storiesKey}'] = 'hi';`;
-
-  const stories = await getStories({ preview: code }, ['preview']);
+  const stories = await getStories({ iframePath: __dirname + '/iframe.html' });
   expect(stories).toEqual('hi');
```
djones: It's less fun.. but consider changing this to `'mock window __storybook_stories__ value`' or similar.

packages/percy-storybook/src/__tests__/getStories-tests.js
timhaines: Yes.  15 seconds wasn't quite long enough either.

PR #46, opened 2018-09-14
Update release notes

maprules1000:
Update release notes to include logging into npm as percy-admin, and the new branch based merge method instead of pushing new versions directly to master.

This process is a bit weird because we're currently using lerna 2.0 (2.11.0 to be specific) to update and publish the packages. It will version them and publish automagically, leaving version bumps that have to be merged into master from a branch. The best solution to this weirdness is to upgrade lerna to v3 so we can run `lerna version`, merge the PR, and then run `lerna publish`.

### DIFF
--- a/RELEASING.MD
+++ b/RELEASING.MD
@@ -1,6 +1,6 @@
 # Releasing
 
-1. check that you are using the percy-admin npm account: `npm whoami`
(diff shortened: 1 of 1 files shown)


Robdel12:
Updated with #182

Issue #50, opened 2018-12-18
Use storybook parameters instead of maintaining a separate store


BPScott:
Currently percy-storybook maintains its own [parameter store](https://github.com/percy/percy-storybook/blob/master/packages/percy-storybook/src/index.js#L46-L62) for attaching parameters to a particular Story.

Storybook 4.0.0 [added support for adding arbitrary parameters to stories](https://github.com/storybooks/storybook/pull/2679).

It would be great if percy-storybook could use this idiomatic way instead of storing data instead of its own store. This would also enable setting global options for all stories within storybook itself instead of having to pass command line flags.

Essentially instead of writing:

```js
 storiesOf('My Story', module)
  .addWithPercyOptions(
    'with multiple widths',
    { widths: [222, 333], rtl: true },
    () => <span>This span renders in widths of 222px and 333px</span>
  );
```

Consumers would be able to write:

```js
storiesOf('My Story', module)
  .add(
    'with multiple widths',
    () => <span>This span renders in widths of 222px and 333px</span>,
    { percy: { widths: [222, 333], rtl: true } }
  );

// OR

storiesOf('My Story', module)
  .addParameters({ percy: { widths: [222, 333], rtl: true } });
  .add(
    'with multiple widths',
    () => <span>This span renders in widths of 222px and 333px</span>,
  );
```

This would also allow global settings defined in storybook config itself instead of having to rely on CLI arguments:

```js
//global parameters, applied to all stories
addParameters({ percy: { widths: [222, 333, 444], rtl: true } });

storiesOf('My Story', module)
  .add(
    'with default widths',
    () => <span>This span renders in widths of 222px, 333px and 444px</span>,
  )
```

Example of parameter usage for reference: https://storybooks-official.netlify.com/?selectedKind=Core%7CParameters&selectedStory=passed%20to%20story&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

marcoala:
Also the `skip` option should use standard parameter api.

timhaines:
This would also allow us to simplify the setup steps for percy-storybook.  

When we do this, I'd suggest percy-storybook moves to the new format, stops supporting the old format, and makes this a major version release.

BPScott:
[...]

PR #64, opened 2019-02-09
Switch to using Storybook's new story parameters

timhaines:
**Important:** This requires a major version bump when releasing and is **dependent on Storybook v5**.

**Breaking Change:** This PR removes [addWithPercyOptions](https://docs.percy.io/docs/storybook-for-react#section-per-story-options) and replaces it with Storybook's own [story parameters](https://github.com/storybooks/storybook/pull/2679) for providing options to Percy.

Options are now provided in a `percy` object in story parameters, either as the 3rd argument to the `add` function, or with a `addParameters` function.  i.e.:

```js
// add Percy parameters to a single story via the 3rd argument to the `add` function.
storiesOf('Circle Stories', module)
  .add(
    'with multiple widths',
    () => <span>This span renders in widths of 222px and 333px</span>,
    { percy: { widths: [222, 333], rtl: true } }
  );

// OR

// add parameters to all Square Stories with addParameters.
storiesOf('Square Stories', module)
  .addParameters({ percy: { widths: [222, 333], rtl: true } });
  .add(
    'with multiple widths',
    () => <span>This span renders in widths of 222px and 333px</span>,
  );

// OR

// add global parameters to all stories with addParameters in your storybook's config.js.
import { configure, setAddon, addParameters } from '@storybook/react';
addParameters({ percy: { widths: [320, 800] } });
```

Storybook's behavior is that adding parameters with the `add` function will be merged into and overwrite clashing parameters provided by `addParameters`.

This PR also significantly simplifies installation, removing the need for the [long step 3](https://docs.percy.io/docs/storybook-for-react#section-installation-https-percy-io-docs-clients-javascript-react-storybook-installation-). 

Closes https://github.com/percy/percy-storybook/issues/50.

Todo:
- [x] Docs update for on each Percy Storybook docs page for this release, simplifying setup, and replacing options section 
- [x] Maintain docs for v2 that supports Storybook < v5.
- [x] Ship as pre-release, and have some folk test it.


timhaines:
To address the caveat, I've submitted a [PR to Storybook](https://github.com/storybooks/storybook/pull/5549) which updates getStorybook to return the parameters.  Let's see how that goes.

If accepted, it will allow us to look at `window['**STORYBOOK_CLIENT_API**'].getStorybook()` instead of `window['**STORYBOOK_CLIENT_API**']['_storyStore']['_data']`

BPScott:
Nice! I got very excited about that `−19,422` diff  before realising it's mostly deleting a stray package.lock, oh well :D

I can't see any tests for global parameters, it'd be nice to make sure those work as you'd expect.

e.g. 

```
//global parameters, applied to all stories
addParameters({ percy: { widths: [222, 333, 444], rtl: true } });

storiesOf('My Story', module)
  .add(
    'with default widths',
    () => <span>This span renders in widths of 222px, 333px and 444px</span>,
  )
```

---

Now that global parameters are possible within Percy, is it worth removing the --widths,  --rtl and --rtl_regex options from the CLI, as they can be easily expressed with a global `addParameters()` call? That might help simplify the docs too as you can talk about CLI options, and then parameters that affect stories that can be configured within Percy.

timhaines:
@BPScott In my one attempt, supplying global params with `addParameters` did not work.  I'll follow up on that.  If they're supposed to work, I'll figure out why they're not.

timhaines:
@BPScott Have confirmed that adding global parameters does work.  My previous test must have been flawed.

Robdel12:
It'd be nice to also switch this package to use the `@percy` namespace on NPM since we're doing a major version bump 

CC: @djones

PR #73, opened 2019-03-16
Remove Lerna - simplifying as there's only one package to publish

timhaines:


Robdel12:
Taken care of with #182

PR #77, opened 2019-04-01
Fix page race condition

loisgomez:
Fix for a possible issue with race condition.
I am getting a page error ```Error: Protocol error (Page.enable): Target closed.```
When running percy-storybook inside a bazel build.

Based on the following issue https://github.com/GoogleChrome/puppeteer/issues/843#issuecomment-331835527


timhaines:
Hi @loisgomez.  Thanks for prepping the PR for this.  Do you have a reproduction of the bug/error we can try this fix against?  I noticed that none of the demo scripts on the puppeteer docs include a separate page.close() step, so would like to confirm it does fix whatever problem you're experiencing before adding this.

Robdel12:
Going to close this out. Please let us know if it's still an issue and we can reopen/explore a fix!

PR #88, opened 2019-04-10
Filter to only relevant attributes when fetching story attributes.

timhaines:
This avoids problems of page.evaluate returning undefined if some of the other story attributes can't be serialized.  ie. when the info for a story contains a jsx fragment.

### DIFF
--- a/packages/percy-storybook/src/getStories.js
+++ b/packages/percy-storybook/src/getStories.js
@@ -16,8 +16,13 @@ const fetchStoriesFromWindow = `(async () => {
     // Usually stories will be found on the first loop.
     var checkStories = function(timesCalled) {
       if (window[storybookClientAPIKey]) {
-        // Found the stories, return them.
-        resolve(window[storybookClientAPIKey].raw());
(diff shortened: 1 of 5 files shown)


packages/percy-storybook/src/getStories.js
```diff
@@ -9,15 +9,18 @@ import { storybookClientAPIKey, storyStoreKey, dataKey } from './constants';
...
     // Usually stories will be found on the first loop.
     var checkStories = function(timesCalled) {
       if (window[storybookClientAPIKey]) {
-        // Found the stories, return them.
-        resolve(window[storybookClientAPIKey].raw());
+        // Found the stories, sanitize to name, kind, and options, and then return them.
+        var reducedStories = window[storybookClientAPIKey].raw().map(story => { return {
+          name: story.name,
+          kind: story.kind,
+          parameters: { percy: story.parameters ? story.parameters.percy : undefined },
```
Robdel12: Really great work here! 🔥

PR #109, opened 2019-06-18
Percy Storybook - add support for minHeight on a per story basis

miguelyoobic95:
- add option to override minimum_height global on a per story basis


Robdel12:
Thank you for this! I'm going to close this PR in favor of #110 so I can push commits to it (for tests & what not). 

Your commits will still be intact 😃

PR #110, opened 2019-06-18
feat: Allow per-story min-height config

Robdel12:
## What is this?

Big thank you to @miguelyoobic95 for taking this on and doing the work. This PR adds the ability to configure per-story min-heights (rather than just a global CLI option).

### DIFF
--- a/integration-tests/storybook-for-react/stories/index.js
+++ b/integration-tests/storybook-for-react/stories/index.js
@@ -111,8 +111,8 @@ storiesOf('addWithPercyOptions', module)
   .add('with skip option', () => <span>Will not Render</span>, {
     percy: { skip: true },
   })
-  .add('with minimum height', () => <span>Renders with multiple</span>, {
-    percy: { minimum_height: 750 },
+  .add('with minimum height', () => <span style={{ height: '20px' }}>Renders with multiple</span>, {
+    percy: { minimum_height: 1500 },
   })
   .add(
     'with RTL of true for a single story',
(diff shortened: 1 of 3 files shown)


packages/percy-storybook/src/selectStories.js
```diff
@@ -50,6 +59,7 @@ export default function selectStories(rawStories, rtlRegex) {
       assertWidths(options.widths);
       assertRtl(options.rtl);
       assertSkip(options.skip);
+      assertMinimumHeight(options.minimum_height);
```
Robdel12: We can bikeshed over snake_case vs camelCase if you want. This currently matches the CLI flag, but I kinda feel like it should match what we have going on for Percy Agent (`minimumHeight` I think?)

integration-tests/storybook-for-react/stories/index.js
```diff
@@ -111,6 +111,9 @@ storiesOf('addWithPercyOptions', module)
   .add('with skip option', () => <span>Will not Render</span>, {
     percy: { skip: true },
   })
+  .add('with minimum height', () => <span style={{ height: '20px' }}>Renders with multiple</span>, {
+    percy: { minimum_height: 1500 },
```
djones:
```suggestion
    percy: { minHeight: 1500 },
```

I think `minHeight` would be more fitting here.

reference: https://github.com/percy/percy-agent/blob/master/src/percy-agent-client/snapshot-options.ts#L5

packages/percy-storybook/src/selectStories.js
djones: ha, yes! `minHeight` :-)

PR #130, opened 2019-08-19
feat: fail when no stories are found

joscha:
About three days ago we upgrade to Storybook v5, unfortunately we didn't notice that the Storybook package didn't correctly pick up the stories any more:

Web__build_UI__78284

This introduces a new CLI flag `--fail_on_empty` which will make `percy-storybook` throw if the stories could not be found.

jest and other unit testing tools fail if their test suite is empty by default (see https://github.com/facebook/jest/issues/6451) for that reason. I personally think this is how Percy should behave as well, but it is theoretically a backwards-incompatible change, so I made the newly introduced flag false by default. Let me know if you are happy to make this a breaking change and I will update the PR.

cc @diagramatics @kongakong

### DIFF
--- a/packages/percy-storybook/src/args.js
+++ b/packages/percy-storybook/src/args.js
@@ -1,6 +1,10 @@
 export const docs = 'Documentation: https://percy.io/percy-storybook';
 
 export const options = {
+  fail_on_empty: {
+    description: 'Fail when no stories are found',
+    requiresArg: true,
+  },
   widths: {
     alias: 'w',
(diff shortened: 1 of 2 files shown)


joscha:
will fix up formatting and tests if the intent of the PR is verified.

joscha:
cc @timhaines

Robdel12:
Hey @joscha this makes sense to me! 👍

joscha:
@Robdel12 changes in!

PR #132, opened 2019-08-20
fix: Yargs wants flags to be strings

Robdel12:
## What is this?

Even though the tests passed for #130, when merged they failed (for a good reason). Looks like yargs wants the default values to be strings rather than any other types. This PR fixes that


joscha:
Ah, thanks!

Robdel12:
No worries, I wouldn't have guessed that 😆

Issue #164, opened 2020-01-06
Regression for test name with single quote character `'`


Marc-Andre-Rivet:
Been using percy storybook for our test suite for over a year now. A test in our test suite started failing last month. The content was always blank.

After investigation / fiddling, renaming the test from `filter override '' & 0` to `filter override empty string & 0` solved the problem (the test isn't blank anymore and contains what's expected -- what's in the storybook-static folder) -- the problematic character seemingly is `'`.

Using `@percy/storybook==3.2.0` and `@storybook/cli==5.1.11` and `@storybook/react==5.1.11`.

Expected resolution would be for either all characters to be correctly supported and/or an appropriate error message describing correct usage if this is intentional, so as not to fail silently.

wwilsman:
Hey @Marc-Andre-Rivet!

Looks like this regression was introduced with Storybook since 5.3. Thanks for bringing it to our attention (and sorry it took so long for us to get to).

Will be fixed once #166 is released 😃

wwilsman:
Took us long enough, but v3.3.0 is now release with this fix. 👍

PR #165, opened 2020-02-03
feat: Use node v10

wwilsman:
_Potentially breaking release_, so tagged as a feature in case users need to pin to an older version rather than update unmaintained node


wwilsman:
Weird snapshot fail.

wwilsman:
Snapshot failure reason: #164

wwilsman:
Ayyyy, passing tests. Thanks #166

PR #166, opened 2020-02-03
fix: Use story id for query format if possible

### LINKED ISSUE #164: Regression for test name with single quote character `'`

wwilsman:
## Purpose

The `selectedKind` and `selectedStory` params are a legacy format that has issues when parameters are url-encoded. Fallback to legacy is still supported.

[Relevant comment from Storybook source](https://github.com/storybookjs/storybook/blob/5c43d0cb449845d5211870a16a141881869cd3a1/lib/core/src/client/preview/url.js#L30-L31)

Fixes #164

## Approach

Use and prefer `story.id` when storing params for selected stories. If not present, the legacy query format is still used.

### Tests

There is already a test for this. It is currently failing for new PRs. As long as all visual tests pass, this fix is verified to work.

### DIFF
--- a/packages/percy-storybook/src/getStories.js
+++ b/packages/percy-storybook/src/getStories.js
@@ -15,11 +15,12 @@ const fetchStoriesFromWindow = `(async () => {
     var checkStories = function(timesCalled) {
       if (window[storybookClientAPIKey]) {
         // Found the stories, sanitize to name, kind, and options, and then return them.
-        var reducedStories = window[storybookClientAPIKey].raw().map(story => { return {
+        var reducedStories = window[storybookClientAPIKey].raw().map(story => ({
+          id: story.id,
           name: story.name,
           kind: story.kind,
           parameters: { percy: story.parameters ? story.parameters.percy : undefined },
-        }});
+        }));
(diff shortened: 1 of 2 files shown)


wwilsman:
🎉 Previously failing snapshot is now passing

Failing: https://percy.io/percy/storybook-for-react/builds/4019975/view/238579704/800?mode=diff&browser=chrome&snapshot=238579704

Passing: https://percy.io/percy/storybook-for-react/builds/4020822/view/238616495/800?mode=diff&browser=chrome&snapshot=238616495

PR #170, opened 2020-02-04
ci: Remove Travis & Appveyor, use GitHub Actions

Robdel12:
## What is this?

This PR removes Travis & Appveyor CI to consolidate into GitHub Actions. A nice benefit of this is we get to dogfood our storybook GH action too.


timhaines:
@Robdel12 Can we maintain testing on windows please?  I think the current state of this PR removes it?  Can we either add on GH Actions, or keep Appveyor.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,42 @@
+name: Windows Tests
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
wwilsman: Is this supposed to be windows? And do we need two workflows for this? Couldn't we have a windows job in the normal ci workflow?

PR #176, opened 2020-02-21
fix: check storybook version number before using id query param

wwilsman:
## Purpose

In #166 preference for `id=${story.id}` was added since the old query format became a legacy format. In that PR, I incorrectly assumed that if a story had an id, then the query parameter would be accepted. That is not the case. Stories do have IDs, but the query param wasn't added until 5.3.x.

## Approach

Rather than prefer the id param when a story has an id, we should check the version of storybook and only use the id param when that version is greater than or equal to 5.3.x.

Rather than add a dependency to check for this, some string parsing is done to grab the major and minor version as numbers. The results of this check is cached so we don't need to perform the parsing for every single story.

### DIFF
--- a/packages/percy-storybook/src/selectStories.js
+++ b/packages/percy-storybook/src/selectStories.js
@@ -1,4 +1,5 @@
 import ExtendableError from 'es6-error';
+import storybookVersion from './storybookVersion';
 
 export class InvalidOptionError extends ExtendableError {}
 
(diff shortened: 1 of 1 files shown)


wwilsman:
Turns out, we use storybook/v5.1.0-alpha.26 in our tests where the id parameter does work. 5.3.x actually just made the other parameters legacy.


Aside: why are we using an older alpha version? Should we update those example apps to use latest stable, or even latest alpha?

PR #187, opened 2020-03-30
feat: retry launching puppeteer when it fails to start

christianscott:
Hey there 👋 puppeteer is failing to launch intermittently on CI with the following error:

```
Error: Failed to launch chrome!
Inconsistency detected by ld.so: dl-tls.c: 493: _dl_allocate_tls_init: Assertion `listp->slotinfo[cnt].gen <= GL(dl_tls_generation)' failed!
TROUBLESHOOTING: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md
    at onClose (/workdir/web/node_modules/puppeteer/lib/Launcher.js:348:14)
    at ChildProcess.<anonymous> (/workdir/web/node_modules/puppeteer/lib/Launcher.js:338:60)
```

Apparently this is a [glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=19329) that's not likely to be resolved. Most of the time it's fixed with a retry. As a workaround we'd like to be able to retry launching puppeteer a certain number of times, so I've added a `puppeteer_launch_retries` argument that defaults to `0`.

### DIFF
--- a/src/args.js
+++ b/src/args.js
@@ -39,7 +39,7 @@ export const options = {
   puppeteer_launch_retries: {
     description: 'Number of times to attempt to launch puppeteer',
     requiresArg: true,
-  }
+  },
 };
 
 export const usage = 'Usage: $0 --widths=320,1280 --debug';
--- a/src/args.js
+++ b/src/args.js
@@ -36,6 +36,10 @@ export const options = {
     description: 'Runs stories with matching names a second time with direction param of rtl',
     requiresArg: true,
   },
+  puppeteer_launch_retries: {
+    description: 'Number of times to attempt to launch puppeteer',
+    requiresArg: true,
+  }
 };
 
 export const usage = 'Usage: $0 --widths=320,1280 --debug';
--- a/src/getStories.js
+++ b/src/getStories.js
@@ -38,7 +38,7 @@ const fetchStoriesFromWindow = `(async () => {
   });
 })()`;
 
-export default async function getStories(options = {}) {
+export default async function getStories(options = {}, launchPuppeteer = opts => puppeteer.launch(opts)) {
   let launchArgs = [];
 
   // Some CI platforms including Travis requires Chrome to be launched without the sandbox
@@ -48,7 +48,7 @@ export default async function getStories(options = {}) {
     launchArgs.push('--no-sandbox');
   }
 
-  const browser = await getBrowser(launchArgs, options.puppeteerLaunchRetries);
+  const browser = await getBrowser(launchPuppeteer, launchArgs, options.puppeteerLaunchRetries);
   const page = await browser.newPage();
 
   await page.goto('file://' + options.iframePath);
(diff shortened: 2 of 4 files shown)


Robdel12:
Hey @christianscott, thanks for the PR here! 😃 I'm curious to know if this issue also exists in your envs with our other SDKs?

christianscott:
Ah, great catch @Robdel12. Yes we're also getting this with `@percy/agent`. If you're happy with this approach & are OK with duplicating the solution across the two packages I can open up a PR on that package too.

christianscott:
Hey @Robdel12! Gentle bump, anything I can do to help get this in?

Robdel12:
Hey @christianscott sorry about the delay here -- I'm not in love with the solve. We've been discussing what to do about this. It may end up being that we have to accept this, but I would rather not patch a _really_ low level bug in our SDKs. I want to learn more about the issue and see if it can be patched lower than our SDKs or not.

Just feels like an odd option to expose and we would have to expose it everywhere we use Puppeteer. 

I will say, if we can't find anything better we'll probably have to to merge. I won't leave you high and dry! But I do want to explore other possible fixes

christianscott:
> I would rather not patch a really low level bug in our SDKs

Totally understand this. It's certainly a weird option to expose.

How would you feel about a custom exit code instead? If we take a step back, the main goal here is to retry when the binary fails to launch due to puppeteer. If we're able to tell when this happens we can implement the retry logic ourselves. A custom exit code would let us do this.

Would be as easy as this:

```typescript
async function getBrowser(args) {
  try {
    return await puppeteer.launch({ headless: true, args });
  } catch (error) {
    console.error(error);
    process.exit(SPECIAL_PUPPETEER_EXIT_CODE);
  }
}
```

My vote would be for 112, ascii code for `p`. Whether that stands for percy or puppeteer is open to interpretation :)

christianscott:
Closing as we haven't seen this error in a while

joscha:
This has just reappeared - any more thoughts @christianscott, @Robdel12 ?

joscha:
this just reappeared in a different form via https://github.com/puppeteer/puppeteer/issues/2207#issuecomment-536517507

@christianscott can you reopen the PR, please?
@Robdel12 any chance we can get this in, please?

Robdel12:
Hey @joscha! The next version of the SDK doesn't use Puppeteer. Is this happeing with 3.x?

joscha:
We're on `@percy/storybook@3.3.1` which seems to depend on `puppeteer "^1.4.0"`

joscha:
Would this be covered by: https://github.com/percy/percy-storybook/issues/236#issuecomment-851111062 ? When would that come out of beta?

joscha:
@Robdel12 ping

Robdel12:
@joscha Pretty soon! We're making lots of improvements to `@percy/core` -- there's a PR coming there and another beta of the storybook SDK. Have you given it a shot by any chance?

joscha:
I'd rather wait until it is out of beta if that is okay ;)

Robdel12:
@joscha no problem! Once this PR ships we should be pulling it out of beta: https://github.com/percy/cli/pull/369

It's stable now, we're just making it faster (for all SDKs, but Storybook and the `snapshot` CLI command are going to be the biggest winners from it)

PR #193, opened 2020-04-06
Pass `puppeteer_launch_args` to `puppeteer.launch` invocation

anulman:
Here we go! Per email thread with @wwilsman


anulman:
Not sure what's up with these unit tests... works on my machine?

image

Robdel12:
Closing this out now that this SDK no longer relies on Puppeteer (it uses the CLI's Chromium browser used in asset discovery). #301

PR #237, opened 2020-09-03
fix: upgrade pupeteer to v5

aviadhahami:
Upgrade puppeteer version


Robdel12:
Closing this out now that this SDK no longer relies on Puppeteer (it uses the CLI's Chromium browser used in asset discovery). #301

PR #301, opened 2021-05-24
✨ Use @percy/cli

wwilsman:
## What is this?

There is a new kid on the block, [`@percy/cli`](https://github.com/percy/cli), that takes what `@percy/agent` has done and makes it more robust, more modular, more extensible, and more feature-rich than `@percy/agent` can ever be in its current state. This PR brings `@percy/cli` to this SDK by refactoring the entire SDK from the ground up, including new tests, new CI config, and other niceties. Since this SDK never used `@percy/agent` to begin with, there are a lot of changes!

### Breaking Changes

- [See the relevant README upgrade section for a list of breaking changes](https://github.com/percy/percy-storybook/blob/ww/new-cli/README.md#upgrading).

### New Features:

- **Support for live URLs and local development!**

  A URL pointing to a live or local Storybook instance can be passed to the CLI command instead of a build directory to take snapshots of the already running Storybook: 
  `$ percy storybook https://my-hosted-storybook.io`.

  During local development, or in CI jobs that don't need to build the full Storybook output, you can also use the sub-command, `$ percy storybook:start`, to automatically run `start-storybook` before taking snapshots and subsequently stop the process after taking snapshots. This sub-command accepts the same arguments as the normal CLI command, plus any additional arguments will be passed directly to the `start-storybook` command:
  `$ percy storybook:start --static-dir ./public`

- **Static snapshots!** 
   
  Currently, this SDK _requires_ JS to be enabled in Percy's infrastructure in order to render stories from the raw preview source and the Storybook JS bundle (which can sometimes be very large, even too large to upload). This can result in flakey screenshots if loading and rendering the story takes a long time, or if JS sometimes causes DOM reflows. 

  With static snapshots, JS is disabled by default and static DOM snapshots are taken of the story. This allows re-renders in Percy's infrastructure to be much more stable and matches all other modern Percy SDKs. The old behavior of a full ground-up render can be re-enabled by enabling JS for a snapshot with the `enableJavaScript` snapshot option.

- **Asset discovery!** 
 
  Previously, this SDK could only upload build-level assets. Along with static snapshots comes asset discovery and per-snapshot resources. Like other SDKs, the static snapshot is loaded into a dedicated headless browser page and network requests are captured when they originate from the Storybook URL, or other allowed hostnames. This enables capture of external but sometimes private or local assets that may not be present in the Storybook build directory.

- **More configuration options!**

  With a new modern foundation, this SDK can now accept the same [Percy config file options](https://docs.percy.io/docs/cli-configuration) as all other SDKs!

  In addition to Percy config file options, per-snapshot options are also supported! In other SDKs, this takes the form of a second argument to the `percySnapshot` method. Since this SDK has no such method, per-snapshot options are accepted as Storybook `percy` parameters. This SDK uses the `percy.capture` method under the hood, so options such as `waitForTimeout` and `waitForSelector` are also accepted.

  [Storybook SDK specific options can be seen in the README configuration section](https://github.com/percy/percy-storybook/blob/ww/new-cli/README.md#configuration)

### Repo Changes:

- Separate GH Action workflows exist for linting and tests so their failures can be reported separately (rather than having tests be cancelled due to failed linting). The test workflow uses a matrix to test on both Linux and Windows.

- The changelog and release processes have also been updated to fully utilize GitHub features such as tags and releases. The semantic release bot and several semantic release packages were replaced with [release-drafter](https://github.com/release-drafter/release-drafter) which utilizes PRs and labels to create changelogs found alongside each release on the releases page.

- Automatic package releasing is now a semi-manual process due to bots not always being able to correctly guess the proper version number and sometimes releasing the incorrect version. To release, the version number should be bumped appropriately and then the automatically created draft release can be published from the releases page. In the future, we can automate the version bump to reflect the release version to make this process smoother, while still requiring a human to approve the release.


bin/percy-storybook
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+
+let log = require('@percy/logger')('percy-storybook-script');
+log.warn('The `percy-storybook` command has been deprecated. Use @percy/cli instead.');
+log.info('See migration instructions here: https://docs.percy.io/docs/migrating-to-percy-cli');
```
Robdel12, line 5: Should this link to a storybook specific migration doc? Seems like it's worth having it's own doc/section

test/storybook-start.test.js
```diff
@@ -0,0 +1,61 @@
...
+      `[percy] Running "start-storybook --host=localhost --port=9000 ${args.join(' ')}"`,
+      '[percy] Percy has started!',
+      '[percy] Created build #1: https://percy.io/test/test/123',
+      '[percy] Found 3 snapshots',
+      '[percy] Snapshot taken: Snapshot: First',
+      '[percy] Snapshot taken: Snapshot: Second',
+      '[percy] Snapshot taken: Skip: But Not Me',
+      '[percy] Finalized build #1: https://percy.io/test/test/123',
+      '[percy] Done!'
+    ]));
```
Robdel12, line 43: 😍

Robdel12 (approved):
🏁 I'm really excited to get this out there. Big jump forward for this SDK! One comment on the deprecation message.

The README looks great too! I do feel like the upgrading section could get a happy path example from old to new.

PR #303, opened 2021-05-27
✨ Automatically add start-storybook --ci arg

wwilsman:
## What is this?

When using `percy storybook:start` we don't want `start-storybook` to open the storybook in a browser or prompt for user input. This adds the `--ci` flag by default to avoid those things.


Robdel12 (approved):
🏁 this makes a lot of sense, nice catch

PR #304, opened 2021-05-28
📝 Refine readme upgrade section

wwilsman:
## What is this?

Adds a small paragraph directly under the upgrade install command that mentions replacing existing usage with new usage. 

Also rewords the breaking changes intro since the recommended upgrade path is to just replace your existing usage. In addition to the rewording, the command change is also listed.


Robdel12 (approved):
🏁 I like it, fresh eyes

PR #305, opened 2021-05-28
📝 Add default min height to breaking changes

wwilsman:
## What is this?

Previous versions of this SDK had a default, flag controlled, minimum height of 800px. In all other SDKs (and now the latest version of this SDK) the default minimum height is handled by the API and our rendering environment, which is set to 1024px. When upgrading and relying on defaults, users may experience height related diffs. This change should be pointed out in the breaking changes section of the readme so people can be informed of the height difference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,9 +174,14 @@ Most importantly, the command itself has changed and all previous arguments are 
   $ percy storybook ./build
   ```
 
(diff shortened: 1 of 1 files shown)


Robdel12 (approved):
🏁 Almost forgot about this

PR #312, opened 2021-06-08
📝  Update README `include` & `exclude` docs

Robdel12:
## What is this?

It's not super obvious these are regex's on the story name (so lets make it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,8 @@ storybook:
   exclude: []
 ```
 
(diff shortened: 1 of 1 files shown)


wwilsman (changes requested):
Let's make these descriptions `An array of regular expressions matching story names...`

wwilsman (approved):
🎮

PR #328, opened 2021-06-24
♻️ Update code to accommodate core changes

wwilsman:
## What is this?

`@percy/core` has some nice recent changes to the internal queue that we need to account for here.

- First, rather than mapping and awaiting on captures, we can synchronously queue them with the improved snapshot method. When `percy.stop()` is eventually called via `finally()`, it will wait for all queued snapshots. When called with an error, it will dequeue all snapshots and attempt to finalize any build immediately.

- The `client.sendSnapshot()` patch needed to be updated to account for a `buildId` parameter.

- Now that the `percy.capture()` method has been merged with `percy.snapshot()`, we should remove a couple of options that could potentially cause issues if provided. Mainly, the `domSnapshot` and `execute` options. We could potentially allow `execute`, but I think it may be confusing that it runs in a different context and won't have access to any local references such as variables or imports.

- The `snapshots` option was renamed to match core's `additionalSnapshots` option. If `snapshots` is provided by a non-skipped story, a single deprecation warning will be logged while collecting snapshots. 

- The `storybook:start` command will now wait for the preview iframe to become available before continuing to start percy. While the main command will always wait for the same thing, it also logs a warning when storybook takes a bit to respond. With the `:start` command, this warning is sometimes logged while storybook is still compiling. The preview is awaited on because I found the root URL to resolve long before storybook is actually ready.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,9 @@
     }
   },
   "dependencies": {
-    "@percy/cli-command": "^1.0.0-beta.52",
-    "@percy/core": "^1.0.0-beta.52",
-    "@percy/logger": "^1.0.0-beta.52",
+    "@percy/cli-command": "^1.0.0-beta.55",
+    "@percy/core": "^1.0.0-beta.55",
+    "@percy/logger": "^1.0.0-beta.55",
     "cross-spawn": "^7.0.3",
     "qs": "^6.10.1",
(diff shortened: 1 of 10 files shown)


Robdel12 (approved):
🏁 Slick! Can't wait to get my hands on this 👏🏼

PR #354, opened 2021-07-30
✨ Add parameters to available config file options

wwilsman:
## What is this?

Before a recent commit to reorganize the readme, it made it seem like all accepted Storybook `percy` parameters were also accepted as config file options. This wasn't the case, as only `include` and `exclude` were accepted, mirroring the accepted command flags. The readme was updated to separate the two sections to better illustrate that Storybook `percy` parameters are _only_ accepted as Storybook parameters.

However, not all config file options need to mirror command flags, and some parameters do make sense to be accepted as config file options. This PR makes this SDK not only accepts relevant Storybook parameters within a Percy config file, but also makes use of new `@percy/config` changes to validate parameters and migrate deprecated options.

This change required a bit of a refactor. As the previous implementation collected all stories, plus `additionalSnapshots`, before filtering through them and managing parameters. Leaving it as such, all options were validated, even for stories that would be skipped. This doesn't seem like a bad design, but it actually made invalid and deprecated options very hard to test since we use a single Storybook as a test harness. The refactor makes it so stories are skipped early, and `additionalSnapshots` are filtered before being mapped to pages and subsequently validated.

Validation warnings still happen during the mapping between Storybook stories and snapshot pages. Because of this, if multiple stories cause warnings, even through inheritance from the global Storybook preview parameters, there would be duplicate warnings one after another; which is less than helpful. Since it is up to this SDK to handle those validation warnings anyway, they are memoized during story collection so they are only ever logged once.

As an aside, it can sometimes be confusing how the `include` and `exclude` parameters are string patterns and real regular expressions do not work. While things were being refactored, regular expressions are now serialized as strings and then transformed back into regular expression during the `shouldSkipStory` test. The schema was also updated to accept RegExps inside JS Percy config files.


Robdel12 (approved):
🏁

PR #358, opened 2021-08-03
🧹 Fix stale issue label & dependabot GH actions

Robdel12:
## What is this?

Forgot to add `stale-issue-label` to the config 🤦🏻‍♂️ (all of the repos added a generic "Stale" label 🤢 )

Also adds dependabot for our actions deps too (to match the CLI config)


wwilsman (approved):
🎮

PR #361, opened 2021-08-05
docs: updated Storybook docs queryParams

iamtyce:
### Description

After testing locally, I confirmed that `queryParams: { globals: 'textDirection:rtl' }` will successfully add a `dir="rtl"` attribute to a story and create a new snapshot for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,13 +112,14 @@ options](https://docs.percy.io/docs/cli-configuration#per-snapshot-configuration
       name: 'My snapshot',
       additionalSnapshots: [
         { prefix: '[Dark mode] ', args: { colorScheme: 'dark' } },
-        { suffix: ' with a search', queryParams: { search: 'foobar' } }
+        { suffix: ' with a search', queryParams: { search: 'foobar' } },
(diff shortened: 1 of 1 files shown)


wwilsman:
Hey @iamtyce! While it appears you can indeed set `globals` via query params like you have there, I don't believe the `textDirection` global is something that comes with Storybook by default. The `dir` attribute is usually something that is handled by the app, and how the app determines that is up to the implementation. Although it's entirely possible there is a Storybook plugin you have installed that handles `globals.textDirection` automatically.

I do think it would be valuable to document the `globals` query param, but it would probably better be documented on [Percy's docs site](https://docs.percy.io/docs) once this SDK leaves beta status. I'm going to close this for now since this doesn't work for all Storybooks, but my current priority is to update the docs site so I can un-beta this SDK properly, so I'll keep the `globals` query param in mind. 🙂

iamtyce:
@wwilsman Makes total sense! Apologies for not looking more into our Storybook setup to realise that we were using our own custom `globals` implementation for text direction. I do still think an example where you can use `globals` would be helpful for other Storybook users but I'll think about a better example to contribute here in future (if at all).

Thanks again, love your work here and we're excited to use Percy!

PR #370, opened 2021-08-12
🔧 Update README & release script for stable release

Robdel12:
## What is this?

This updates the README to remove the `@next` install instructions and also removes the `--next` flag from the release script.


wwilsman (changes requested):
I believe there is a `prerelease` (or similar) option in the release drafter config that needs to be removed as well.

wwilsman (approved):
I see now that it was already removed by accident in another PR. Ship away

🎮

PR #371, opened 2021-08-13
🔇 Remove deprecated `snapshots` option & logs

Robdel12:
## What is this?

The deprecated `snapshots` option (now `additionalSnapshots`) no longer needs to hang around now that we're on a stable release.


wwilsman (approved):
🎮 🎉

PR #394, opened 2021-09-02
🐛 Gather Storybook version for SDK UA

Robdel12:
## What is this?

Currently we're not collecting the version of Storybook when creating builds. This makes it a little harder to provide support (and becomes a question to the customer rather than us just having it). 

The version is missing from v4 because how we gathered the version in the past was a bit cumbersome. It would look at the dependency tree and try to match a storybook framework version. There's no single package for us to import to get the version from. 

The way we grab the version now is by visiting the `?path=/settings/about` route and plucking the version from the DOM. Storybook uses this route to show the current version & does a version check. 

A small improvement was made to how we get gather the stories off of the `window` object. We now use the `waitFor` util to double check the var is present _before_ trying to capture the stories. Helpful for remote storybooks that take a bit longer to load/render.

Probably should merge/release this before this CLI PR (to avoid warnings) https://github.com/percy/cli/pull/479

### DIFF
--- a/src/command.js
+++ b/src/command.js
@@ -11,6 +11,12 @@ import pkg from '../package.json';
 
 // used to deserialize regular expression strings
 const RE_REGEXP = /^\/(.+)\/(\w+)?$/;
+const WAIT_FOR_TIMEOUT = 5000;
+
+// remove stack traces from any eval error
+function normalizeEvalError(error) {
+  return error.replace(/^Error:\s(.*?)\n\s{4}at\s.*$/s, '$1');
+}
 
 export default class StorybookCommand extends Command {
   static flags = {
@@ -48,7 +54,7 @@ export default class StorybookCommand extends Command {
     });
 
     let url = await this.storybook();
-    // borrow a browser page to discover stories
+    // borrow a browser page to get the storybook version and discover stories
     await this.percy.browser.launch();
     let [version, pages] = await Promise.all([this.getStorybookVersion(url), this.getStoryPages(url)]);
     let l = pages.length;
@@ -192,22 +198,16 @@ export default class StorybookCommand extends Command {
       await page.goto(previewUrl);
 
       /* istanbul ignore next: no instrumenting injected code */
-      return await page.eval(async ({ waitFor }, previewUrl) => {
+      return await page.eval(async ({ waitFor }, waitForTimeout) => {
         // ensure the page has loaded and the var we need is present
-        await waitFor(() => !!window.__STORYBOOK_CLIENT_API__, 5000)
+        await waitFor(() => !!window.__STORYBOOK_CLIENT_API__, waitForTimeout)
           .catch(() => Promise.reject(new Error(
             'Storybook object not found on the window. ' +
             'Open Storybook and check the console for errors.'
           )));
+
         let serializeRegExp = r => r && [].concat(r).map(r => r.toString());
         let storybook = window.__STORYBOOK_CLIENT_API__;
-
-        if (!storybook) {
-          throw new Error(
-            'Storybook object not found on the window. ' +
-            'Open Storybook and check the console for errors.'
-          );
-        }
 
         return storybook.raw().map(({ id, kind, name, parameters }) => ({
           name: `${kind}: ${name}`,
@@ -216,11 +216,11 @@ export default class StorybookCommand extends Command {
           exclude: serializeRegExp(parameters?.percy?.exclude),
           id
         }));
-      });
+      }, WAIT_FOR_TIMEOUT);
     } catch (error) {
       // remove stack traces from any eval error
       if (typeof error === 'string') {
-        throw new Error(error.replace(/^Error:\s(.*?)\n\s{4}at\s.*$/s, '$1'));
+        throw new Error(normalizeEvalError(error));
       } else {
         throw error;
       }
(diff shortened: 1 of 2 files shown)


src/command.js
```diff
@@ -185,13 +188,17 @@ export default class StorybookCommand extends Command {
...
 
       /* istanbul ignore next: no instrumenting injected code */
-      return await page.eval((util, previewUrl) => {
+      return await page.eval(async ({ waitFor }, previewUrl) => {
+        // ensure the page has loaded and the var we need is present
+        await waitFor(() => !!window.__STORYBOOK_CLIENT_API__, 5000)
+          .catch(() => Promise.reject(new Error(
+            'Storybook object not found on the window. ' +
+            'Open Storybook and check the console for errors.'
+          )));
```
wwilsman, line 201: Now that we have this check & error, we don't need to do the same check & error below.

Also, small style thing, can we add a newline below this while your making the above change?

src/command.js
```diff
@@ -221,4 +228,31 @@ export default class StorybookCommand extends Command {
...
   }
+
+  async getStorybookVersion(url) {
+    let version, page;
+
+    try {
+      page = await this.percy.browser.page();
+      this.log.debug(`Get Storybook version: ${url}/?path=/settings/about`);
+
+      await page.goto(`${url}/?path=/settings/about`);
```
wwilsman, line 239: We should construct this URL like we do other URLs, like so:

```js
let aboutUrl = new URL('?path=/settings/about', url);
```

It serves to both validate the provided URL, and ensure we aren't accidentally creating an invalid or incorrect URL (such as double `//` in the middle, or if there are existing URL fragments)

src/command.js
```diff
@@ -221,4 +228,31 @@ export default class StorybookCommand extends Command {
...
+      /* istanbul ignore next: no instrumenting injected code */
+      version = await page.eval(async ({ waitFor }) => {
+        await waitFor(() => !!document.querySelector('header'), 5000)
+          .catch(() => Promise.reject(new Error('Failed to find a <header> element')));
+        let getPath = path => document.evaluate(path, document, null, 9, null).singleNodeValue;
```
wwilsman, line 244: Could move `getPath` to before the wait, then utilize it when waiting so we wait on the exact element we're looking for.

src/command.js
```diff
@@ -221,4 +228,31 @@ export default class StorybookCommand extends Command {
...
+        let text = getPath("//header[starts-with(text(), 'Storybook ')]").innerText;
+
+        return text.match(/[-]{0,1}[\d]*[.]{0,1}[\d]+/g, '').join('');
+      });
+    } catch (error) {
+      this.log.debug(`Couldn't retrieve Storybook version: ${error.message}`);
```
wwilsman, line 250: In the matching eval pattern for story collection, we remove ugly stack traces from any error messages. We can move that regexp into a constant, and/or make a small helper function to handle eval errors.

test/storybook.test.js
```diff
@@ -176,6 +176,16 @@ describe('percy storybook', () => {
...
   });
 
+  it('sends the version of storybook when creating snapshots', async () => {
+    await Storybook.run(['http://localhost:9000']);
+
+    expect(logger.stderr).toEqual([]);
+    expect(mockAPI.requests['/builds/123/snapshots'].map(req => req.headers['user-agent'])[0])
+      .toMatch(
+        /^Percy\/v1 @percy\/client\/\S+ @percy\/storybook\/\d+\.\d+\.\d+.*? \(storybook\/\d+\.\d+\.\d+.*?; node\/v[\d.]+.*\)$/
+      );
```
wwilsman, line 186: I don't think we need to necessarily test the entire user-agent. Just that it contains what we intend it to.

```suggestion
      .toMatch(/storybook\/\d+\.\d+\.\d+.*?;/);
```

src/command.js
```diff
@@ -48,9 +48,12 @@ export default class StorybookCommand extends Command {
     });
 
     let url = await this.storybook();
-    let pages = await this.getStoryPages(url);
+    // borrow a browser page to discover stories
```
wwilsman, line 51: Can probably update this comment

```suggestion
    // borrow a browser page to get the storybook version and discover stories
```

src/command.js
```diff
@@ -185,13 +188,17 @@ export default class StorybookCommand extends Command {
...
-      // borrow a browser page to discover stories
-      await this.percy.browser.launch();
       page = await this.percy.browser.page({ meta });
       await page.goto(previewUrl);
 
       /* istanbul ignore next: no instrumenting injected code */
-      return await page.eval((util, previewUrl) => {
+      return await page.eval(async ({ waitFor }, previewUrl) => {
+        // ensure the page has loaded and the var we need is present
+        await waitFor(() => !!window.__STORYBOOK_CLIENT_API__, 5000)
```
wwilsman, line 197: Should the `5000` timeout be a constant at the top, shared with both evals?

We can pass the constant inside of eval by adding another arg, like `previewUrl` gets.... that arg does nothing. So you can rename that arg to `timeout` and pass in the constant as a second arg to `eval`.

wwilsman (changes requested):

src/command.js
```diff
@@ -11,6 +11,12 @@ import pkg from '../package.json';
 
 // used to deserialize regular expression strings
 const RE_REGEXP = /^\/(.+)\/(\w+)?$/;
+const WAIT_FOR_TIMEOUT = 5000;
+
+// remove stack traces from any eval error
+function normalizeEvalError(error) {
+  return error.replace(/^Error:\s(.*?)\n\s{4}at\s.*$/s, '$1');
```
wwilsman, line 18: In case a real error is thrown, we should add a string check like the other catch has for this:

```suggestion
  if (typeof error !== 'string') return error;
  return new Error(error.replace(/^Error:\s(.*?)\n\s{4}at\s.*$/s, '$1'));
```

src/command.js
```diff
@@ -209,16 +216,48 @@ export default class StorybookCommand extends Command {
...
-      });
+      }, WAIT_FOR_TIMEOUT);
     } catch (error) {
       // remove stack traces from any eval error
       if (typeof error === 'string') {
-        throw new Error(error.replace(/^Error:\s(.*?)\n\s{4}at\s.*$/s, '$1'));
+        throw new Error(normalizeEvalError(error));
       } else {
         throw error;
       }
```
wwilsman, line 226: With the above change, this can then become

``` javascript
  throw normalizeEvalError(error);
```

src/command.js
```diff
@@ -209,16 +216,48 @@ export default class StorybookCommand extends Command {
...
+        await waitFor(() => getPath(headerPath), waitForTimeout)
+          .catch(() => Promise.reject(new Error('Failed to find a <header> element')));
+
+        return getPath(headerPath)
+          .innerText
+          .match(/[-]{0,1}[\d]*[.]{0,1}[\d]+/g, '')
+          .join('');
+      }, WAIT_FOR_TIMEOUT);
+    } catch (error) {
+      this.log.debug(`Couldn't retrieve Storybook version: ${normalizeEvalError(error)}`);
```
wwilsman, line 255: To handle real errors that aren't strings, this can become:

```suggestion
      this.log.debug('Unable to retrieve Storybook version');
      this.log.debug(normalizeEvalError(error));
```

(the `debug` method will handle any real stack traces)

src/command.js
```diff
@@ -209,16 +216,48 @@ export default class StorybookCommand extends Command {
...
         throw error;
       }
     } finally {
       await page?.close();
     }
   }
+
+  async getStorybookVersion(url) {
+    let version, page;
+    let aboutUrl = new URL('?path=/settings/about', url).href;
```
wwilsman, line 234: Small style thing I wasn't going to comment on until I made the other comments. Can we list uninitialized variables last? (version and page under the url var)

src/command.js
```diff
@@ -209,16 +216,48 @@ export default class StorybookCommand extends Command {
...
+      /* istanbul ignore next: no instrumenting injected code */
+      version = await page.eval(async ({ waitFor }, waitForTimeout) => {
+        let headerPath = "//header[starts-with(text(), 'Storybook ')]";
+        let getPath = path => document.evaluate(path, document, null, 9, null).singleNodeValue;
+
+        await waitFor(() => getPath(headerPath), waitForTimeout)
+          .catch(() => Promise.reject(new Error('Failed to find a <header> element')));
+
+        return getPath(headerPath)
+          .innerText
```
wwilsman, line 250: Another small style thing I'm lumping in with the other only cause I already made comments. 😆 

Can `.innerText` stay inline with `getPath` since it is the subject of the method chain below.

src/command.js
```diff
@@ -11,6 +11,12 @@ import pkg from '../package.json';
 
 // used to deserialize regular expression strings
 const RE_REGEXP = /^\/(.+)\/(\w+)?$/;
+const WAIT_FOR_TIMEOUT = 5000;
```
wwilsman, line 14: Last small pedantic thing, I promise. 

I feel like this name is a little vague (wait for what?). How about `EVAL_TIMEOUT`?

wwilsman (changes requested):
One main suggestion around error handling (what if you catch an error that is not a string?)

The remaining comments are small things I only added cause I was making a suggestion anyway.

wwilsman (approved):
🎮

PR #558, opened 2022-05-05
docs(README): improve documentation for Puppeteer options

oscard0m:


wwilsman:
While these options have the same name and behavior as Puppeteer, we do not actually use Puppeteer.

oscard0m:
> While these options have the same name and behaviour as Puppeteer, we do not actually use Puppeteer.

Thanks @wwilsman 

I'm opening an issue where we can discuss how to improve documentation on those then.

PR #611, opened 2022-07-20
🐛 Add missing tests and miscellaneous fixes

wwilsman:
## What is this?

With #592, it was discovered that this SDK was actually missing quite a bit of tests. Since it was out of scope for that PR, we opted to address this in another PR (this PR).

With the addition of new tests to achieve 100% coverage, a few small bugs inevitably came up and were fixed.

- When splitting snapshots into shards with `--shard-count`, providing the `--partial` flag will now always set the parallel total environment variable to `-1` unless already set. Previously, the `--shard-count` flag did not consider the `--partial` flag when setting the environment variable.

- When creating the story URL and settings params with the native URLSearchParams API, it will encode any provided values. However, args and globals params are already encoded, therefore resulting in double encoding of those params. This was fixed with string concatenation for already encoded values, while still using the URLSearchParams API for provided query params.

- Both the `storybook` and `storybook:start` commands have similar try-catch blocks to start Percy and call the same utility function with a different `baseUrl`. The try-catch repetition was moved into the utility function which now accepts a callback that is called during the `finally` block.

- The `deferUploads` option was changed to the new `delayUploads` option so snapshots get sent to the API as soon as possible rather than all at once at the end of the process. Uploads are still delayed rather than be left eager so empty builds are not created when failing to find any snapshots or Storybook build.

- A new event listener was added to the end of story navigation, which will wait until Storybook emits a `storyRendered` event before taking a snapshot. While network idle is always awaited on, this might help in situations when no network requests occur, which may cause the snapshot to be taken too early when no other wait options are provided.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,5 @@
 import { request, createRootResource } from '@percy/cli-command/utils';
+import { buildArgsParam } from '@storybook/router';
 
 // Transforms authorization credentials into a basic auth header and returns all config request
 // headers with the additional authorization header if not already set.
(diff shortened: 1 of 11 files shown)


src/utils.js
```diff
@@ -115,10 +101,19 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
   ))).then(channel => {
+    // emit a series of events to render the desired story
     channel.emit('updateGlobals', { globals: {} });
     channel.emit('setCurrentStory', { storyId: story.id });
     channel.emit('updateQueryParams', { ...story.queryParams });
     channel.emit('updateGlobals', { globals: story.globals });
     channel.emit('updateStoryArgs', { storyId: story.id, updatedArgs: story.args });
+
+    // resolve when rendered, reject on any other renderer event
+    return new Promise((resolve, reject) => {
```
Robdel12, line 112: Slicccckk

Robdel12 (approved):
🏁 Pretty uncharacteristic for us to miss coverage like this but very happy to have this SDK back up to 100%.

PR #619, opened 2022-07-26
♻️ Remove dependency on storybook param util

wwilsman:
## What is this?

We need to be able to recreate acceptable Storybook query params for args and globals when generating a story's URL. To do this, we have historically relied on Storybook's own internal utils to ensure 1:1 compatibility.

Unfortunately, this dependency has proven to be highly volatile. Early minor updates had the util moving between files, while other updates effected the package's build output. Most recently, The package's distributed files have become tightly bundled to React. While this doesn't present a problem when used within a React Storybook project, it does however cause dependency errors when used outside of React environments (even when the target Storybook is using React).

With these changes, we no longer depend on a Storybook internal package, and instead reimplement and recreate Storybook URL params as closely as possible. The encode/decode functions introduced here were copied and adapted from Storybook's internal util to fit our own coding style. I suspect (and hope) that future updates will only require adding additional features rather than changing existing encoding/decoding behavior.

While making these changes, I realized that certain values were being lost between the transfer from the browser execution context to Node's execution context. To preserve these values, all args and globals are now serialized and encoded before being transferred to the Node context. This is done by providing page eval scripts with additional Storybook specific helper functions. Any args and globals specified within a Percy config file are also encoded separately to ensure uniformity when building each story snapshot's URL.

An additional story snapshot was added as well as several unit tests to add coverage for the new appropriated Storybook params utils. Some error reporting was also adjusted to ease debugging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,9 @@
     ]
   },
   "dependencies": {
-    "@percy/cli-command": "^1.6.3",
-    "@storybook/router": "^6.5.0",
-    "cross-spawn": "^7.0.3"
+    "@percy/cli-command": "^1.6.4",
+    "cross-spawn": "^7.0.3",
+    "qs": "^6.11.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.18.6",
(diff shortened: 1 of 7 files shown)


Robdel12 (approved):
🏁 hopefully this makes it so this SDK is no longer a full time job to maintain anymore. Probably should watch their repo now to keep track of any changes, eh?

PR #622, opened 2022-07-29
🐛 Fix process aborting

wwilsman:
## What is this?

Currently, `Ctrl+C` or normal kill signals do not work to stop the Storybook SDK process from running. This is because the majority of the work is done in promises, which block CLI's abort handling.

This PR fixes this issue by updating `@percy/cli-command` and refactoring a few promises utilizing new generator utils.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     ]
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


Robdel12 (approved):
🏁

PR #625, opened 2022-08-03
🐛 Fix capture options validation warnings

### LINKED ISSUE #621: Incorrect invalid option warning for `waitForSelector`
### Discussed in https://github.com/percy/percy-storybook/discussions/618

<div type='discussions-op-text'>

<sup>Originally posted by **JodiWarren** July 26, 2022</sup>
Within our `.percy.yml` we have the `waitForSelector` option set like so:

```yaml
storybook:
  waitForSelector: '#root > *'
```

When running percy, we receive this error on each snapshot:
```
[percy] Invalid snapshot options:
[percy] - waitForSelector: not accepted with DOM snapshots
```

According to the docs this seems a valid configuration: https://docs.percy.io/docs/storybook#percy-config-file-options
</div>

---

It looks like this is a bug with recent changes. The snapshot option is accepted, but this message is being incorrectly logged after the option is already used when taking the DOM snapshot to begin with.

We should strip these capture options after they have been used before they are passed along with the DOM.

wwilsman:
## What is this?

This PR fixes #621 by utilizing validation helpers to scrub unused snapshot options without warning about them. A test was adjusted slightly to ensure the warnings are no longer logged when capture options such as `waitForSelector` are provided.

While making this change, I realized that these capture options are not actually used at all when JavaScript is enabled. This is intended since enabling JavaScript tells this SDK to use the raw preview DOM to avoid rehydration issues. But it may be confusing if some people expect these options to work when they won't be used at all. 

To address this confusion, the config schema was refactored to validate against capture options when JavaScript is enabled. This required a couple small upstream changes and a much more complex schema for this SDK. The result is when JavaScript is enabled in for a snapshot (via a config file, storybook parameters, snapshot options, etc), capture options are validated against to warn that they are not used with JavaScript enabled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     ]
   },
   "dependencies": {
-    "@percy/cli-command": "^1.7.2",
+    "@percy/cli-command": "^1.8.0",
     "cross-spawn": "^7.0.3",
     "qs": "^6.11.0"
   },
--- a/src/config.js
+++ b/src/config.js
@@ -1,10 +1,35 @@
+import { snapshotSchema } from '@percy/core/config';
+
 export const storybookSchema = {
   $id: '/storybook',
   $ref: '/storybook#/$defs/params',
   $defs: {
+    enableJavaScript: {
+      isTrue: {
+        required: ['enableJavaScript'],
+        properties: { enableJavaScript: { const: true } }
+      },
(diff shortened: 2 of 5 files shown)


Robdel12 (approved):
🏁

PR #651, opened 2022-09-21
🐛 Preemptively update usage of internal core methods

wwilsman:
## What is this?

A few internal core methods will be changing in the next CLI update and will require some changes to this SDK for it to continue working correctly. These changes can be made preemptively before the next CLI release since they will continue to work with older CLI versions.


Robdel12 (approved):
🏁

PR #717, opened 2023-03-13
Use @percy/cli from github for regression

samarsault:
Support trigger workflows via dispatch from CLI https://github.com/percy/cli/pull/1201 for regression

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,11 @@ on:
     branches: [master]
   pull_request:
   workflow_dispatch:
+    inputs:
+      branch:
+        required: false
+        type: string
+        default: master
 jobs:
   test:
     name: Test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,4 +26,18 @@ jobs:
         env:
           PERCY_POSTINSTALL_BROWSER: true
       - run: yarn build
+      - name: Set up @percy/cli from git
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: |
+          cd /tmp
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -26,4 +26,18 @@ jobs:
         env:
           PERCY_POSTINSTALL_BROWSER: true
       - run: yarn build
+      - name: Set up @percy/cli from git
```
itsjwala, line 29: make `on:` similar to other repos. 

```yml
on: 
  push:
  workflow_dispatch:
    inputs:
      branch:
        required: false
        type: string
        default: master
```

itsjwala: is `L28` - `yarn build` required? I believe `yarn test:coverage` handles building as well?

itsjwala (changes requested):

.github/workflows/test.yml
samarsault: I think we need it

.github/workflows/test.yml
```diff
@@ -4,6 +4,11 @@ on:
     branches: [master]
   pull_request:
   workflow_dispatch:
+    inputs:
```
itsjwala, line 7:
```suggestion
  workflow_dispatch:
    inputs:
```

.github/workflows/test.yml
itsjwala: can you try it once? since other repos doesn't have it

samarsault: Removed it

itsjwala (approved):

PR #723, opened 2023-03-23
CLI regression fixes

samarsault:
In our pipeline we link @percy/cli here as part of all packages. It is not used and creates some resolution issues. Fixes that

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,7 @@
 name: Test
 on:
   push:
-    # branches: [master]
+    branches: [master]
   workflow_dispatch:
     inputs:
       branch:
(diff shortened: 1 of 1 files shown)


samarsault:
Window CI related PRs - https://github.com/percy/percy-storybook/pull/9, https://github.com/percy/percy-storybook/pull/170
I guess this was more relevant when percy-storybook was independent. Since we test the CLI now on both, we can remove it

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     name: Test
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest]
```
itsjwala, line 16: verify once if we rely on paths or any OS specific code base lets remove if gtg

samarsault: verified

test/storybook.test.js
```diff
@@ -191,7 +192,7 @@ describe('percy storybook', () => {
       req.body.data.relationships.resources
         .data.find(r => r.attributes['is-root']).id
     ]).sort((a, b) => a[0] > b[0] ? 1 : -1)).toEqual([
-      ['foo: bar', sha(storyDOM)],
+      ['foo: bar', sha(storyDOM.replace('<head></head>', `<head>${polyfill}</head>`))],
```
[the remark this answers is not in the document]
samarsault: `percy.snapshot` received the html with injection. We can't have an assertion for the original HTML content since there's no reference to it. CLI goes to the page and gives us the HTML. Let us keep it this way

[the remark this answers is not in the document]
samarsault: Added

test/storybook.test.js
```diff
@@ -174,6 +173,10 @@ describe('percy storybook', () => {
     // respond with the preview dom only for the first request
     server.reply('/iframe.html', () => [200, 'text/html', i++ ? storyDOM : previewDOM]);
 
+    // eslint-disable-next-line import/no-extraneous-dependencies
+    let { Percy } = await import('@percy/core');
```
itsjwala, line 177: lets import on top of the file

samarsault: It's done this way in other tests as well. We don't want spy to propagate

itsjwala (approved):

PR #819, opened 2023-10-12
✨  Skip story on error and move to next stories

ninadbstack:
Context:
- Currently if percySnapshot fails on one of the stories [ say dom.js returned an exception or timedout ] we throw that error again
- This breaks test suite and no further stories are uploaded

Change:
- We add an env var `PERCY_SKIP_STORY_ON_ERROR` which if set to 'true' would log the error skip the story and move to next story 
- We also recreate the page to avoid any state related issues

### DIFF
--- a/src/snapshots.js
+++ b/src/snapshots.js
@@ -188,6 +188,14 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags 
     // set storybook environment info
     percy.client.addEnvironmentInfo(environmentInfo);
 
+    // We use an outer and inner loop on same snapshots.length
+    // - we create a new page and load one story on it at a time for snapshotting
+    // - if it throws exception then we want to catch it outside of `withPage` call as
+    //   when `withPage` returns it closes the page
+    // - we want to make sure we close the page that had exception in story to make sure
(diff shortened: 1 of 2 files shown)


src/snapshots.js
```diff
@@ -188,38 +188,54 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
-      lastCount = snapshots.length;
-
-      while (snapshots.length) {
-        // separate story and snapshot options
-        let { id, args, globals, queryParams, ...options } = snapshots[0];
-
-        if (flags.dryRun || options.enableJavaScript || percy.config.snapshot.enableJavaScript) {
-          // when dry-running or when javascript is enabled, use the preview dom
-          options.domSnapshot = previewResource.content;
+    while (snapshots.length) {
```
itsjwala, line 191: nit: add a comment here why we're not reusing the same page.

itsjwala (commented):
lgtm

itsjwala (approved):

src/snapshots.js
ninadbstack: Added

PR #829, opened 2023-10-18
✨ add storybook v7 support

nilshah98:
**Description -**
- [x] update `evalStorybookEnvironmentInfo` to grep storybook v7 version
- [x] update `evalStorybookStorySnapshots` to support fetching all stories even with `storyStorev7: true`
- [x] update packages to storybook v7 and update tests (`start-storybook` to `storybook dev`)
- [x] update minimum node version required to `16` and update github actions to use node `16` 
- [x] verify `cacheAllCSFFiles` with `storyStorev7: false` and also older storybook version
- [x] debug flaky tests


src/utils.js
```diff
@@ -180,6 +183,9 @@ export function evalStorybookStorySnapshots({ waitFor }) {
   };
 
   return waitFor(async () => {
+    // uncache stories, if cached via storyStorev7: true
```
itsjwala, line 186: does this work with `storyStorev7: false` ?

itsjwala: Lets add a test for the same.

src/start.js
```diff
@@ -32,11 +32,11 @@ export const start = command('start', {
   let { default: { spawn } } = yield import('cross-spawn');
   let { host, port } = flags;
 
-  let args = ['--ci', `--host=${host}`, `--port=${port}`, ...argv];
-  log.info(`Running "start-storybook ${args.join(' ')}"`);
+  let args = ['dev', '--ci', `--host=${host}`, `--port=${port}`, ...argv];
```
itsjwala, line 35: we'll need backward compatibility with v6.

itsjwala (commented):
Looks good so far, nice!

* we'll need backward compatibility with v6 and have to figure out how we can add tests for same.
* we'll create E2E build on Percy and verify visual regression if any with v6 vs v7 and callout.
  * how many customers use node 14? we might need to have CI tests on node 14 as well.

package.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/percy/percy-storybook.git"
   },
   "engine": {
-    "node": ">=14"
+    "node": ">=16"
```
itsjwala, line 10: we don't need to have this compulsion, our code will still work with node 14.

(though does it let us run just verify) ?

nilshah98: Nope, storybook v7 is [**not** compatible with node 14](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#dropped-support-for-node-15-and-below).

Screenshot 2023-10-19 at 9 49 29 PM

src/start.js
nilshah98: Yep, this is a little tricky.
I am unsure how to structure the SDK in that case.

src/utils.js
nilshah98: Will check this.
Also need to verify this with older versions of storybook

nilshah98:
I have noticed that the tests are flaky.
I am not sure if this was the behaviour before as well, might need to debug this.
Example -
[Successful run](https://github.com/percy/percy-storybook/actions/runs/6577372147)
[Unsuccessful run](https://github.com/percy/percy-storybook/actions/runs/6577372147/job/17868869143?pr=829)

Both of them are for the same commit ids

src/utils.js
nilshah98: Refer - https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#v7-store-api-changes-for-addon-authors

test/storybook.test.js
```diff
@@ -49,7 +55,11 @@ describe('percy storybook', () => {
   it('snapshots live urls', async () => {
     await storybook(['http://localhost:9000']);
 
-    expect(logger.stderr).toEqual([]);
+    // if there are stderr logs ensure it is only an acceptable warning
+    expect(logger.stderr).toEqual(logger.stderr.length ? [
+      '[percy] Waiting on a response from Storybook...'
```
itsjwala, line 60:
```suggestion
      '[percy] Waiting on a response from Storybook...'
```

why was this required? are we not waiting enough for storybook?

itsjwala (commented):
looks good, some customary changes are pending.

apart from what we discussed in sync.

src/utils.js
itsjwala: v6 and v7 with `true` tests are covered, can you add tests for v7 with `false`

test/storybook.test.js
nilshah98: We have added extra await for [checkStorybookVersion](https://github.com/percy/percy-storybook/blob/add-storybookv7-support/test/storybook.test.js#L20), this might be causing delays for storybook to start.
Hence, needed to add extra waiting check.

We already do something similar in [storybook-start.test.js](https://github.com/percy/percy-storybook/blob/master/test/storybook-start.test.js#L25)

src/utils.js
nilshah98: This will make the tests even more complicated.
We _might_ have to run all test cases again.

Why is this an issue, as this has been officially documented by storybook team, that we need to call `cacheAllCSFFiles` as documented above.

src/utils.js
```diff
@@ -147,7 +162,10 @@ export async function* withPage(percy, url, callback, retry) {
 // Evaluate and return Storybook environment information from the about page
 /* istanbul ignore next: no instrumenting injected code */
 export function evalStorybookEnvironmentInfo({ waitForXPath }) {
-  return waitForXPath("//header[starts-with(text(), 'Storybook ')]", 5000)
+  let possibleEnvs = [];
```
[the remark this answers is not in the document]
nilshah98: I think current tests are good enough.
We are already checking for `storybook/<number>`.

Current production is throwing `storybook/unknown` for storybook v7.

src/utils.js
itsjwala: we don't want a case where calling when its `false` and it errors out.

Though I'll keep this requirement optional if this is tested locally by creating a percy build.

src/utils.js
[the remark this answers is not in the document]
itsjwala: I meant check for equality of this xpath in the test case and if it returns correct version.
or another option is when posting the snapshot we intercept the network call to verify the environment details being sent.

I its already there we can skip, this is to avoid regression when storybook is updated locally and its go untested.

itsjwala (approved):
LGTM 🔥

src/utils.js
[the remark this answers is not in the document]
nilshah98: There is already test for this at -
https://github.com/percy/percy-storybook/blob/master/test/storybook.test.js#L245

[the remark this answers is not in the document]
itsjwala: This is resolved in sync, there is already a test for this.

PR #834, opened 2023-10-27
🔖  update package version to 5.0.0-beta.1

nilshah98:


packageV6.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@percy/storybook",
-  "version": "4.3.7",
+  "version": "5.0.0-beta.1",
```
itsjwala, line 3: `"private": true` make this private so we won't publish this file, and we can remove version information from here.

itsjwala (approved):

PR #837, opened 2023-10-31
🐛 ✨Fix eval setCurrentStory error + make logging better

nilshah98:
**Description-**
- [x] Fixed logging, earler we were just doing `reject()` without passing any error, hence it was passed as `undefined` and hence error was `undefined` and we were unable to parse `message` from it
- [x] Fixed issue where loading `iframe.html` as base preview URL would trigger the `storyMissing` event which would call `reject` and break flow. Solution for this was to open first component, instead of `/iframe.html`

### DIFF
--- a/src/snapshots.js
+++ b/src/snapshots.js
@@ -199,7 +199,9 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags 
     while (snapshots.length) {
       try {
         // use a single page to capture story snapshots without reloading
-        yield* withPage(percy, previewUrl, async function*(page) {
+        // loading an existing story instead of just iframe.html as that triggers `storyMissing` event
+        // This in turn leads to promise rejection and failure
+        yield* withPage(percy, `${previewUrl}?id=${snapshots[0].id}&viewMode=story`, async function*(page) {
           // determines when to retry page crashes
           lastCount = snapshots.length;
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -142,7 +142,7 @@ export async function* withPage(percy, url, callback, retry) {
     return yield* yieldTo(callback(page));
   } catch (error) {
     // if the page crashed and retry returns truthy, try again
-    if (error?.message?.includes('crashed') && retry?.()) {
+    if (error.message?.includes('crashed') && retry?.()) {
       return yield* withPage(...arguments);
     }
 
@@ -247,9 +247,9 @@ export function evalSetCurrentStory({ waitFor }, story) {
     // resolve when rendered, reject on any other renderer event
     return new Promise((resolve, reject) => {
       channel.on('storyRendered', resolve);
-      channel.on('storyMissing', reject);
-      channel.on('storyErrored', reject);
-      channel.on('storyThrewException', reject);
+      channel.on('storyMissing', () => reject('Story Missing'));
+      channel.on('storyErrored', () => reject('Story Errored'));
+      channel.on('storyThrewException', () => reject('Story Threw Exception'));
     });
   });
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -247,9 +247,9 @@ export function evalSetCurrentStory({ waitFor }, story) {
     // resolve when rendered, reject on any other renderer event
     return new Promise((resolve, reject) => {
       channel.on('storyRendered', resolve);
-      channel.on('storyMissing', () => reject(new Error('Story Missing')));
-      channel.on('storyErrored', () => reject(new Error('Story Errored')));
-      channel.on('storyThrewException', () => reject(new Error('Story Threw Exception')));
+      channel.on('storyMissing', (err) => reject(err || new Error('Story Missing')));
+      channel.on('storyErrored', (err) => reject(err || new Error('Story Errored')));
+      channel.on('storyThrewException', (err) => reject(err || new Error('Story Threw Exception')));
     });
   });
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -247,9 +247,9 @@ export function evalSetCurrentStory({ waitFor }, story) {
     // resolve when rendered, reject on any other renderer event
     return new Promise((resolve, reject) => {
       channel.on('storyRendered', resolve);
-      channel.on('storyMissing', () => reject('Story Missing'));
-      channel.on('storyErrored', () => reject('Story Errored'));
-      channel.on('storyThrewException', () => reject('Story Threw Exception'));
+      channel.on('storyMissing', () => reject(new Error('Story Missing')));
+      channel.on('storyErrored', () => reject(new Error('Story Errored')));
+      channel.on('storyThrewException', () => reject(new Error('Story Threw Exception')));
     });
   });
 }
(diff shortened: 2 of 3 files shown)


src/utils.js
```diff
@@ -247,9 +247,9 @@ export function evalSetCurrentStory({ waitFor }, story) {
     // resolve when rendered, reject on any other renderer event
     return new Promise((resolve, reject) => {
       channel.on('storyRendered', resolve);
-      channel.on('storyMissing', reject);
-      channel.on('storyErrored', reject);
-      channel.on('storyThrewException', reject);
+      channel.on('storyMissing', () => reject(new Error('Story Missing')));
```
ninadbstack, line 250: Do we get original error passed to us

test/storybook.test.js
```diff
@@ -156,12 +156,12 @@ describe('percy storybook', () => {
 
     await expectAsync(storybook(['http://localhost:8000']))
     // message contains the client stack trace
-      .toBeRejectedWithError(/^Story Error\n.*\/iframe\.html.*$/s);
+      .toBeRejectedWithError(/^Story Errored\n.*\/iframe\.html.*$/s);
```
ninadbstack, line 159: Word change ? Is this same for v6 and v7

src/utils.js
nilshah98: Nope, there are no args passed to this event.
I tried -
```js
channel.on('storyMissing', (...args) => console.log(args))
```
But this didn't log anything.

test/storybook.test.js
nilshah98: This is a custom error that we are throwing.
Earlier in the test itself, we were mocking the error to be thrown, since `storyErrored` event was caught, but we `reject` the promise and passed nothing to it, so it returned undefined and fell back to the mocked error.

src/snapshots.js
```diff
@@ -199,7 +199,9 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
     while (snapshots.length) {
       try {
         // use a single page to capture story snapshots without reloading
-        yield* withPage(percy, previewUrl, async function*(page) {
+        // loading an existing story instead of just iframe.html as that triggers `storyMissing` event
+        // This in turn leads to promise rejection and failure
+        yield* withPage(percy, `${previewUrl}?id=${snapshots[0].id}&viewMode=story`, async function*(page) {
```
itsjwala, line 204: can you refactor this bit and pull out `snapshots[0]` out its used inside as well.

Also one question, 
* why this was not an issue till now?

src/snapshots.js
```diff
@@ -199,7 +199,9 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
     while (snapshots.length) {
       try {
         // use a single page to capture story snapshots without reloading
-        yield* withPage(percy, previewUrl, async function*(page) {
+        // loading an existing story instead of just iframe.html as that triggers `storyMissing` event
```
itsjwala, line 202: one more refactoring change not related to current PR,
we had made were having `await` on the promise 
https://github.com/percy/percy-storybook/blob/254b3266ea3e645c87eab7b3430564cf30a49526/src/start.js#L35


instead we'll use yield, this is to support cancellation of promise.

src/utils.js
nilshah98: Even tried -
```js
      channel.on('storyMissing', (err) => {
        reject(err);
      });
```
But this again resulted in -
`[percy] TypeError: Cannot read properties of undefined (reading 'message')`

nilshah98: Even tried -
```js
      channel.on('storyMissing', (err) => {
        reject(err);
      });
```
But this again resulted in -
`[percy] TypeError: Cannot read properties of undefined (reading 'message')`

nilshah98: Oh, just learned something new about JS which explains this better.
Whenever a function name is passed as callback, it calls the function by whatever parameters are passed.
So, in this case
```js
channel.on('storyMissing', reject)
```

This would call reject with whatever parameters are passed to eventListener, and since nothing is passed, hence we reject with `undefined` resulting in above error.

src/snapshots.js
nilshah98: I am not 100% sure about this.
My guess is, there are some additional functions in Storybook v7 which were not present in v6.
Example - https://github.com/storybookjs/storybook/blob/v7.5.0/code/lib/preview-api/src/modules/preview-web/PreviewWithSelection.tsx#L491
This file is not present in `v6.5.16`

src/snapshots.js
nilshah98: I tried rewriting this using generators, but I am unable to do os.
I am unable to find any good examples, the existing solution that I checked out on stackoverflow involved creating an internal array, where resolved promises push values to, and the generator yields from there.
https://stackoverflow.com/questions/51045136/how-can-i-use-a-event-emitter-as-an-async-generator

Open to suggestion if there is any other way to refactor this.

src/utils.js
nilshah98: > so basically story missing isnt passing anything ... 

yes

> Also its not a new thing, its a standard thing that works across almost all languages ( except where brackets are not required for a function call, Ruby 👀 ), it works because its no different, it expects a function and you are passing a function

Wasn't aware of this, thanks!

test/storybook.test.js
nilshah98: Refactored error throwing logic, so as to not update tests.

src/snapshots.js
itsjwala: Alright, let me check the details, not a blocker. 

can you address other comments.

>can you refactor this bit and pull out snapshots[0] out its used inside as well.

> evalSetCurrentStory we use this already to change the story, can you test with multiple story examples (mostly test case should have this scenario, just confirm)

src/snapshots.js
itsjwala: changing `await` to `yield` works?

src/snapshots.js
nilshah98: > can you refactor this bit and pull out snapshots[0] out its used inside as well.

This wouldn't be very beneficial, as I will still need to extract `id` and other parameters from `snapshots[0]` over [here](https://github.com/percy/percy-storybook/blob/master/src/snapshots.js#L208), since it is inside a while loop, and can have different `snapshots[0]` than what `withPage` is called with.

> evalSetCurrentStory we use this already to change the story, can you test with multiple story examples (mostly test case should have this scenario, just confirm)

Tested locally and verified.

ninadbstack (approved):
LGTM

src/snapshots.js
itsjwala: not a blocker for this release.

let me refactor it alongside some other PR.

src/snapshots.js
itsjwala: > This wouldn't be very beneficial, as I will still need to extract id and other parameters from snapshots[0] over [here](https://github.com/percy/percy-storybook/blob/master/src/snapshots.js#L208), since it is inside a while loop, and can have different snapshots[0] than what withPage is called with.


that's not correct, outer while loop is only meant if the page crashes or something, and we're iterating on the same snapshot array so anyway inner while loop will consume all the snapshot array.

not a blocker for this release, we can maybe refactor in future PRs.

itsjwala (approved):

PR #845, opened 2023-11-27
feat: add captureServiceWorker flag when opening browser

nilshah98:
**context -**
- we have added support for capturing service worker intercepted requests in https://github.com/percy/cli/pull/1443
- in msw, it checks if service worker controller is registered or not. If it is not registered then it goes into an infinite rendering loop.
- when using percy browser, by default we have `setByPassServiceWorker: false` which does not allow registration of service worker controller and hence we go in infinite rendering loop.
- adding `captureServiceWorker: true` enables us to register service worker controller, and hence resolves the infinite rendering loop.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -116,7 +116,7 @@ export async function* withPage(percy, url, callback, retry) {
   let page = yield percy.browser.page({
     networkIdleTimeout: percy.config.discovery.networkIdleTimeout,
     requestHeaders: getAuthHeaders(percy.config.discovery),
-    captureMockedServiceWorker: percy.config.captureMockedServiceWorker,
+    captureMockedServiceWorker: percy.config.discovery.captureMockedServiceWorker,
     userAgent: percy.config.discovery.userAgent
   });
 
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -116,6 +116,7 @@ export async function* withPage(percy, url, callback, retry) {
   let page = yield percy.browser.page({
     networkIdleTimeout: percy.config.discovery.networkIdleTimeout,
     requestHeaders: getAuthHeaders(percy.config.discovery),
+    captureServiceWorker: percy.config.captureServiceWorker,
```
itsjwala, line 119: Note: this would mean we're using global value here, and in the CLI PR we're allowing `per-snapshot level config.

itsjwala: So per snapshot level doesn't make sense for storybook SDK, since we'll need to create fresh new page. 

so we'll have to remove per-snapshot level support from CLI or we can keep it, but it will create confusion.

ninadbstack (approved):

PR #847, opened 2023-12-01
feat: added defer uploads for percy-storybook

nilshah98:
**Context-**
- `defer-uploads` is used for capturing multiple dom snapshots, for same webpage.
- but in percy-storybook, we do not respect the `defer-uploads` config
- this pr, consumes `defer-uploads` config, and then takes multiple snapshots for same webpage depending on number of widths

**Why test cases have changes so much ?**
- Earlier percy-storybook versions were using percy cli version - `1.24.0`
- In this version, the `defer-uploads` config was not passed successfully from cli -> storybook due to configSchema being outdated in `1.24.0`
- Hence, updated `cli-command` package to `1.27.4` which fixed this issue and `defer-uploads` tests started working.
- But with `1.27.4` we also updated the [createTestServer](https://github.com/percy/cli/blob/master/packages/core/test/helpers/server.js#L4) mock service with [this](https://github.com/percy/cli/pull/1387/files) particular PR.
- In summary what above PR does is, it treats `/iframe.html` and `/iframe.html?id=1` as different requests and responds to them differently.
- Hence, in percy storybook, we first load `iframe.html` get all stories and then navigate to `iframe.html?id={storyId}` which is not available anymore.
- Therefore added additional mocks, which return same response for `iframe.html` and `iframe.html?id={storyId}&viewMode=story` route.

**Related PRs-**
- https://github.com/percy/cli/pull/1450

### DIFF
--- a/src/snapshots.js
+++ b/src/snapshots.js
@@ -231,7 +231,7 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags 
                   options.domSnapshot = domSnapshot;
                   // validate without logging to prune all other options
                   PercyConfig.validate(options, '/snapshot/dom');
-                  // validate without logging to prune all other options
+                  // snapshots are queued and do not need to be awaited on
                   percy.snapshot({ ...options, widths: [w] });
                 }
               } else {
@@ -239,7 +239,7 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags 
                 options.domSnapshot = domSnapshot;
                 // validate without logging to prune all other options
                 PercyConfig.validate(options, '/snapshot/dom');
-                // validate without logging to prune all other options
+                // snapshots are queued and do not need to be awaited on
                 percy.snapshot({ ...options });
               }
             }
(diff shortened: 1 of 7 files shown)


src/snapshots.js
```diff
@@ -213,19 +213,27 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
+              // snapshots are queued and do not need to be awaited on
+              percy.snapshot(options);
             } else {
               log.debug(`Loading story: ${options.name}`);
               // when not dry-running and javascript is not enabled, capture the story dom
               yield page.eval(evalSetCurrentStory, { id, args, globals, queryParams });
               /* istanbul ignore next: tested, but coverage is stripped */
-              let { dom, domSnapshot = dom } = yield page.snapshot(options);
-              options.domSnapshot = domSnapshot;
+              for (let i = 0; i < percy.config.snapshot.widths.length; i++) {
```
itsjwala, line 225: This will increase User's CI time considerably, we've seen few customers using on packed CI with multiple widhts.

we should ideally ship this behind feature, or have it as accepted bug.

Issue #911, opened 2024-03-26
When upgrading to Storybook 8 I get "Error: Storybook object not found on the window. Open Storybook and check the console for errors."


rudeayelo:
## The problem

When I try tu upgrade to the latest version of Storybook (`8.0.4` currently) and I trigger the `percy storybook` command, the process stops just after downloading Chromium.

## Environment

- Node version: `v20.5.1`
- `@percy/cli` version: `1.28.2`
- `@percy/storybook` version: `5.0.1`
- Version of Percy SDK you’re using: -
- If needed, a build or snapshot ID: -
- OS version: MacOS Sonoma 14.4 (23E214)
- Type of shell command-line [interface]: `zsh`

## Details

If necessary, describe the problem you have been experiencing in more detail.

## Debug logs

```
[percy] Successfully downloaded Chromium 929475
[percy] Percy has started!
[percy] Stopping percy...
[percy] Build not created
[percy] Error: Storybook object not found on the window. Open Storybook and check the console for errors.
error: script "test-storybook:visual" exited with code 1
```

## Code to reproduce issue

I grabbed the React starter template from Storybook, upgraded it and added Percy, same issue appears, here's the repo: https://github.com/rudeayelo/percy-storybook-8

ArnaudWeyts:
Also running into this, but it seems like it's flaky for me. Sometimes it passes, sometimes it throws this error.

JackHowa:
Experienced this too. It sometimes works?

jalandis:
Just ran into this issue while upgrading to Storybook 8.

```
[percy:cli] Error: Storybook object not found on the window. Open Storybook and check the console for errors.
```

jonthomp:
The official news from support is that Percy/Browserstack are not planning on fixing this until at least September

dkremez:
We are running into the same issue.
```
[percy] Error: Storybook object not found on the window. Open Storybook and check the console for errors.
error Command failed with exit code 1.
```

rudeayelo:
I reported it through BrowserStack support center, as suggested here: https://github.com/percy/percy-storybook/pull/918#issuecomment-2027162812

MaelFr:
Hello,
Received some news regarding this issue:
> The Engineering team … informed me that they have prioritized this task and begun working on the Storybook upgrade for v8. We anticipate completing this upgrade by the 20th of May (Tentative).

rishigupta1599:
Hey,
A quick update here.
[...]

PR #918, opened 2024-03-28
Add Storybook 8 compatibility

MaelFr:
Hello,

I faced [this issue](https://github.com/percy/percy-storybook/issues/911), which seems to be related to [this part of the migration](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#storystore-and-methods-deprecated).
I made this change and tried it locally. It worked.
It should be better to add the [dependabot PR to upgrade to Storybook 8](https://github.com/percy/percy-storybook/pull/908), but the node version in tests should be changed to 18+.
Additionnaly, I was unable to run the tests due to download failure (maybe related to my company's security policy) :

> ✗ snapshots live urls
      - Error: Download failed: 404 - https://storage.googleapis.com/chromium-browser-snapshots/Mac_Arm/929475/chrome-mac.zip

There are definitely a lot more things that need to change to fully handle v8, but this seems like a good start.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -200,12 +200,14 @@ export function evalStorybookStorySnapshots({ waitFor }) {
 
   return waitFor(async () => {
     // uncache stories, if cached via storyStorev7: true
-    await (window.__STORYBOOK_PREVIEW__?.cacheAllCSFFiles?.() ||
+    await (window.__STORYBOOK_PREVIEW__?.ready?.() ||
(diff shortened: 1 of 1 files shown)


ninadbstack:
@MaelFr Thanks for debugging the issue and for initial PR. As this will need full scale support and testing before we can merge it, I will recommend you to create a support ticket with Browserstack and we would pick this accordingly!

PR #993, opened 2024-07-17
Add retry in withPage function

pankaj443:
- Adding retry in withPage function.
- Update evalSetCurrentStory swap:  `setCurrentStory` & `updateGlobals`.

Above changes are with respect to storybook v8 fixes, we were getting `execution context destroyed` on several stories. swapping above calls fixes this.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -139,7 +139,7 @@ export function decodeStoryArgs(value) {
 
 // Borrows a percy discovery browser page to navigate to a URL and evaluate a function, returning
 // the results and normalizing any thrown errors.
-export async function* withPage(percy, url, callback, retry) {
+export async function* withPage(percy, url, callback, retry, args) {
   let log = logger('storybook:utils');
   let attempt = 0;
   let retries = 3;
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -138,52 +139,66 @@ export function decodeStoryArgs(value) {
...
+          // keep the stack trace if the error came from a client script
+          /\n\s+at\s.+?\(https?:/.test(error) ? '$1' : '$2'
+        )));
+      } finally {
+        // always clean up and close the page
+        await page?.close();
+      }
+    } catch (error) {
+      attempt++;
+      let shouldRetry = process.env.ENABLE_RETRY === 'true';
```
rishigupta1599, line 196: Should we name something specific ENABL_RETRY_ON_STORY_ERROR something like this?, not sure we might add these kind of retry mechanism in CLI or somewhere else in future, so might confuse that time.

rishigupta1599, line 196: By default we wanted env variable to be true but above will require it to explicitly getting set, just confirm once?

pankaj443: done

rishigupta1599 (approved):
LGTM

PR #1072, opened 2024-11-22
updating error message and fixing retry snapshot bug

prklm10:
updating error message and fixing retry snapshot bug

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -193,12 +193,12 @@ export async function* withPage(percy, url, callback, retry, args) {
     } catch (error) {
       attempt++;
       let enableRetry = process.env.PERCY_RETRY_STORY_ON_ERROR || 'true';
+      const from = args?.from;
       if (!(enableRetry === 'true') || attempt === retries) {
         // Add snapshotName to the error message
         const snapshotName = args?.snapshotName;
-        const errorMessage = args?.errorMessage;
-        if (errorMessage) {
-          error.message = `${errorMessage}: \n${error.message}`;
+        if (from) {
+          error.message = `${from}: \n${error.message}`;
         }
         if (snapshotName) {
           error.message = `Snapshot Name: ${snapshotName}: \n${error.message}`;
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -195,9 +195,14 @@ export async function* withPage(percy, url, callback, retry, args) {
...
       if (!(enableRetry === 'true') || attempt === retries) {
+        // Add snapshotName to the error message
+        const snapshotName = args?.snapshotName;
+        if (snapshotName) {
+          error.message = `${error.message} - Snapshot Name: ${snapshotName}`;
+        }
         throw error;
       }
-      log.warn(`Retrying Story: ${args.snapshotName}`);
+      log.warn(`Retrying Story: ${args?.snapshotName}`);
```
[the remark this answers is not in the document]
prklm10: Updated

src/utils.js
```diff
@@ -195,9 +195,15 @@ export async function* withPage(percy, url, callback, retry, args) {
       attempt++;
       let enableRetry = process.env.PERCY_RETRY_STORY_ON_ERROR || 'true';
       if (!(enableRetry === 'true') || attempt === retries) {
+        // Add snapshotName to the error message
+        const snapshotName = args?.snapshotName;
+        if (snapshotName) {
+          error.message = `${error.message} - Snapshot Name: ${snapshotName}`;
```
ninadbstack, line 201: Shouldnt this be Snapshot name followed by error.message ? wouldnt that make it easy to understand snapshot in case of long error messages ?

ninadbstack: Also the original comment is not handled. THe message should be passed from outside. Considering "Retrying Story" makes no sense if we are opening preview page

prklm10: Passed form inside try when attempt number > 0.

test/storybook.test.js
```diff
@@ -173,12 +173,15 @@ describe('percy storybook', () => {
 
     await expectAsync(storybook(['http://localhost:8000']))
     // message contains the client stack trace
-      .toBeRejectedWithError(/^Story Error\n.*\/iframe\.html.*$/s);
+      .toBeRejectedWithError(/^Snapshot Name:/s);
```
ninadbstack, line 176: This is incorrect change

prklm10: fixed. Snapshot name is getting appended to the error.

ninadbstack (approved):

PR #1136, opened 2025-05-22
Added check to capture snapshot after loader

bs-shobhitkumar:


src/utils.js
```diff
@@ -312,10 +312,109 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+          const preparingDocs = document.querySelector('.sb-preparing-docs');
+          const bodyClasses = document.body.classList.toString();
+
+          let loaderInfo = {};
+          if (initialLoader) loaderInfo.initialLoader = true;
+          if (preparingStory) loaderInfo.preparingStory = !!window.getComputedStyle(preparingStory).display !== 'none';
+          if (preparingDocs) loaderInfo.preparingDocs = !!window.getComputedStyle(preparingDocs).display !== 'none';
+          loaderInfo.bodyClasses = bodyClasses;
+
+          reject(new Error(`Timed out waiting for Storybook loaders to disappear: ${JSON.stringify(loaderInfo)}`));
```
ninadbstack, line 346: we shouldnt be rejecting right ? we want to print warning and take snapshot anyway after timeout

ninadbstack (approved):

PR #1154, opened 2025-07-08
Upgrade Storybook to V9

bhokaremoin:
Changes - 
1. Updated the Node Version to 20, which is required for Supporting Storybook V9 - https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#nodejs--20
2. Moved existing package.json to packageV8.json and created a new package.json with changes required for storybook v9

### DIFF
--- a/.github/workflows/Semgrep.yml
+++ b/.github/workflows/Semgrep.yml
@@ -40,12 +40,6 @@ jobs:
         env:
             # Add the rules that Semgrep uses by setting the SEMGREP_RULES environment variable. 
             SEMGREP_RULES: p/default # more at semgrep.dev/explore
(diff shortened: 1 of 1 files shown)


packageV8.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "@percy/storybook",
+  "version": "6.0.4",
```
pankaj443, line 3: this should in 8 right?

bhokaremoin: https://github.com/percy/percy-storybook/blob/master/package.json#L3
That is the percy-sdk version that was released.

pankaj443: ack

PR #1169, opened 2025-07-29
Handling responsiveSnapshotCapture from Build and Snapshot Level

rahulXbrowserstack:


src/common.js
```diff
@@ -12,6 +12,18 @@ export const flags = [{
...
   short: 'e'
+}, {
+  name: 'responsive-snapshot-capture',
+  description: 'Enable responsive DOM capture for multiple viewport widths',
+  percyrc: 'snapshot.responsiveSnapshotCapture',
+  type: 'boolean',
+  default: false
+}, {
+  name: 'widths',
+  description: 'Comma-separated list of viewport widths for responsive capture (e.g. 375,768,1280)',
```
bhokaremoin, line 23: check in current percy-storybook sdk what are default width, keep that same.

From CLI its this - https://github.com/percy/cli/blob/56d0fe4722eea63cfe8a6ea7aa02cf00d465a85d/packages/core/README.md?plain=1#L46
once check with percy-sdk

src/config.js
```diff
@@ -42,6 +42,18 @@ export const storybookSchema = {
...
+          type: 'boolean',
+          default: false
+        },
+        widths: {
+          type: 'array',
+          items: {
+            type: 'integer',
+            minimum: 1
+          },
+          default: []
```
bhokaremoin: shouldn't default value of widths be added here ?

src/snapshots.js
```diff
@@ -8,9 +8,80 @@ import {
...
+    return domSnapshot;
+  }
+}
+
+// Capture responsive DOM snapshots
+async function* captureResponsiveDOM(page, story, options, flags, enableJavaScript, previewResource, percy, log) {
+  if (flags.dryRun || enableJavaScript) {
+    return options?.widths?.map(width => ({
+      width,
+      html: previewResource.content.html || previewResource.content,
```
bhokaremoin: when javascript is enabled, why are we sending only single dom.
1. Is this flow dev tested
2. this is one of the mail flow, if we are sending single dom, please mention here in comments, why

src/snapshots.js
```diff
@@ -8,9 +8,80 @@ import {
...
+      cookies: [],
+      resources: [],
+      hints: [],
+      userAgent: 'percy-storybook',
+      warnings: []
+    }));
+  } else {
+    // Always do actual responsive capture when not in dry-run mode
```
bhokaremoin: add few debug logs

src/snapshots.js
```diff
@@ -8,9 +8,80 @@ import {
...
+    mobile: Array.isArray(deviceDetails) ? deviceDetails.map(d => d.width).filter(Boolean) : [],
+    config: percy.config.snapshot?.widths || []
+  };
+
+  // Get widths regardless of responsive capture mode
+  const widths = getWidthsForResponsiveCapture(
+    options.widths,
+    eligibleWidths
+  );
+
```
bhokaremoin: move all of this inside the if block, we don't want to make this request and calculate widths if `responsiveSnapshotCapture` is false

src/snapshots.js
```diff
@@ -8,9 +8,80 @@ import {
...
+  if (responsiveSnapshotCapture) {
+    const responsiveOptions = { ...options, responsiveSnapshotCapture: true, widths };
+    return yield* captureResponsiveDOM(page, story, responsiveOptions, flags, enableJavaScript, previewResource, percy, log);
+  } else {
+    const singleDOMOptions = { ...options, widths };
```
bhokaremoin, line 49: add debug logs

src/utils.js
```diff
@@ -1,6 +1,10 @@
 import { request, createRootResource, yieldTo } from '@percy/cli-command/utils';
 import { logger } from '@percy/cli-command';
 import spawn from 'cross-spawn';
+// Maximum pixel height for screenshots to prevent excessive memory usage.
+// This limit is set to 25,000 pixels based on practical browser and system constraints.
+const CS_MAX_SCREENSHOT_LIMIT = 25000;
```
bhokaremoin: `CS_MAX_SCREENSHOT_LIMIT` max limit is not required. they don't use storybook sdk.

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+
+  // Remove duplicates
+  return [...new Set(allWidths)].filter(e => e);
+}
+
+// Check if responsive snapshot capture is enabled with proper hierarchy
+export function isResponsiveSnapshotCaptureEnabled(options, config) {
+  // Always disable if defer uploads is enabled
+  if (config?.percy?.deferUploads) {
+    return false;
```
bhokaremoin: I don't think this will be required, as here user is not calling percySnapshot command for each story. Once check code implementation of CLI around this. If required mention why or else remove it

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+}
+
+async function changeViewportDimensionAndWait(page, width, height, resizeCount, log) {
+  try {
+    // Use Percy's CDP-based resize method
+    await page.resize({
+      width,
+      height,
+      deviceScaleFactor: 1,
+      mobile: false
```
bhokaremoin, line 462: check once if we need to update deviceScaleFactor for mobile widths, cross check with selenium sdk

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+    });
+  } catch (e) {
+    log.debug('Resizing using CDP failed, falling back to page eval for width', width, e);
+    // Fallback to JavaScript execution
+    await page.eval(({ width, height }) => {
+      window.resizeTo(width, height);
+      // Trigger resize events and force layout
```
bhokaremoin: wrap this also in try catch and raise error if this fails

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+
+    await new Promise(resolve => setTimeout(resolve, scrollSleep * 1000));
+
+    // Recalculate scroll height for dynamically loaded pages
+    scrollHeight = await page.eval(scrollHeightCommand);
+  }
+
+  // Get back to top
+  await page.eval(() => {
+    window.scrollTo(0, 0);
```
bhokaremoin: why is this function required, are this changes required for responsiveCapture ? I don't think so, check once

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+        domSnapshot.cookies = [];
+      }
+    }
+
+    return domSnapshot;
+  } catch (error) {
+    log.error('Error in captureSerializedDOM:', error);
+    throw new Error(`Failed to capture DOM snapshot: ${error.message}`);
+  }
+}
```
bhokaremoin: why do we have two implementation of captureSerializedDOM function. one in utils and other in snapshot.js.

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+    if (!domSnapshot || typeof domSnapshot !== 'object' || !domSnapshot.html) {
+      throw new Error('Invalid DOM snapshot structure - missing html content');
+    }
+
+    // Add cookies if not already present
+    if (!domSnapshot.cookies) {
+      try {
+        domSnapshot.cookies = await page.eval(() => {
+          if (!document.cookie) return [];
+          return document.cookie.split(';').map(cookie => {
```
bhokaremoin: also why is this cookie logic added here ? and why is it only added in captureSerializedDOM function which is used in captureResponsiveStoryDOM, does the original captureSerializedDOM does not require this ?

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+  }));
+
+  currentWidth = viewportSize.width;
+  currentHeight = viewportSize.height;
+
+  let lastWindowWidth = currentWidth;
+  let resizeCount = 0;
+
+  // Ensure PercyDOM is available before starting
+  yield page.insertPercyDom();
```
bhokaremoin: page.insertPercyDom(); this not required at the end we are doing page.snapshot, this is already handled there

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+      yield page.insertPercyDom();
+    }
+
+    if (process.env.RESPONSIVE_CAPTURE_SLEEP_TIME) {
+      await new Promise(resolve => setTimeout(resolve, parseInt(process.env.RESPONSIVE_CAPTURE_SLEEP_TIME) * 1000));
+    }
+
+    if (process.env.PERCY_ENABLE_LAZY_LOADING_SCROLL) {
+      await slowScrollToBottom(page);
+    }
```
bhokaremoin: is this change required if we want to do responsiveCapture true

test/utils.test.js
```diff
@@ -0,0 +1,79 @@
...
+  it('returns true if config.snapshot.responsiveSnapshotCapture is true', () => {
+    expect(isResponsiveSnapshotCaptureEnabled({}, { snapshot: { responsiveSnapshotCapture: true } })).toBe(true);
+  });
+
+  it('returns false if config.snapshot.responsiveSnapshotCapture is false', () => {
+    expect(isResponsiveSnapshotCaptureEnabled({}, { snapshot: { responsiveSnapshotCapture: false } })).toBe(false);
+  });
+
+  it('returns false if nothing is set', () => {
+    expect(isResponsiveSnapshotCaptureEnabled({}, {})).toBe(false);
```
bhokaremoin: where is the test case where snapshot option will take precedence over global option for responsiveSnapshotCapture flag

test/utils.test.js
```diff
@@ -0,0 +1,79 @@
...
+  const storybookSchema = {
+    responsiveSnapshotCapture: { type: 'boolean', default: false },
+    widths: { type: 'array', items: { type: 'integer', minimum: 1 }, default: [] }
+  };
+  it('responsiveSnapshotCapture default is false', () => {
+    expect(storybookSchema.responsiveSnapshotCapture.default).toBe(false);
+  });
+  it('widths default is [] and type is array of int', () => {
+    expect(storybookSchema.widths.default).toEqual([]);
+    expect(storybookSchema.widths.items.type).toBe('integer');
```
bhokaremoin: there are many other new functions are have been added to utils.js, we need test for those as well

test/snapshots.test.js
```diff
@@ -0,0 +1,88 @@
...
+    await gen.next(); // run to yield percy.client.getDeviceDetails
+    await gen.next(); // run to yield page.eval
+    let result = await gen.next(); // run to yield page.snapshot
+    // Should return the domSnapshot string (handle both string and object)
+    if (typeof result.value === 'object' && result.value.domSnapshot) {
+      expect(result.value.domSnapshot).toBe('<html>real</html>');
+    } else {
+      expect(result.value).toBe('<html>real</html>');
+    }
+  });
```
bhokaremoin, line 84: major test cases are missing for captureDom function based on responsiveSnapshotCapture flag what is the result. for both responsiveSnapshotCapture - true and false

test/snapshots.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+  it('returns array of doms in dryRun', async function() {
+    let flags = { dryRun: true };
+    let options = { widths: [375, 800], responsiveSnapshotCapture: true };
+    let gen = captureResponsiveDOM(page, { id: 'id' }, options, flags, false, previewResource, percy, log);
+    let result = await gen.next();
+    expect(Array.isArray(result.value)).toBe(true);
+    expect(result.value.length).toBe(2);
+    expect(result.value[0].width).toBe(375);
+    expect(result.value[1].width).toBe(800);
```
bhokaremoin: what is its not dryRun, write test for that cases also

test/snapshots.test.js
```diff
@@ -0,0 +1,88 @@
...
+  });
+});
```
bhokaremoin: also test cases for enable_javascript true / false with responsiveSnapshotCapture true / false

src/utils.js
```diff
@@ -417,3 +421,244 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+    );
+  };
+
+  let scrollHeight = Math.min(await page.eval(scrollHeightCommand), CS_MAX_SCREENSHOT_LIMIT);
+  const clientHeight = await page.eval(() => document.documentElement.clientHeight);
+  let current = 0;
+
+  let pageNum = 1;
+  // Break the loop if maximum scroll height 25000px is reached
+  while (scrollHeight > current && current < CS_MAX_SCREENSHOT_LIMIT) {
```
bhokaremoin: ask product if this max scrollHeight is required.

bhokaremoin: i don't think

bhokaremoin (changes requested):

src/snapshots.js
rahulXbrowserstack: if you will see the previous code which was there, it was doing the same, so I didn't changed any thing here, just put that inside a function. Here's the same previous code - https://github.com/percy/percy-storybook/blob/833d07be059d51a950cea04683beab816596b7ae/src/snapshots.js#L212

src/snapshots.js
rahulXbrowserstack: sure

src/utils.js
rahulXbrowserstack: its the same in selenium one , checked it

src/snapshots.js
```diff
@@ -217,9 +260,7 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
               log.debug(`Loading story: ${options.name}`);
               // when not dry-running and javascript is not enabled, capture the story dom
               yield page.eval(evalSetCurrentStory, { id, args, globals, queryParams });
-              /* istanbul ignore next: tested, but coverage is stripped */
-              let { dom, domSnapshot = dom } = yield page.snapshot(options);
-              options.domSnapshot = domSnapshot;
+              options.domSnapshot = await captureDOM(page, options, percy, log);
```
bhokaremoin, line 263: with enableJavaScript true we are not checking for responsiveSnapshotCapture ?

src/snapshots.js
bhokaremoin: yes but we need handling that with enable javascript also we are capturing dom on all widths.

src/utils.js
bhokaremoin: was this checked and then removed ?, if yes please ack it and mention why we removed it

test/utils.test.js
```diff
@@ -0,0 +1,147 @@
+import * as utils from '../src/utils.js';
+
+describe('getWidthsForResponsiveCapture', () => {
+  it('merges mobile and user widths, then adds config defaults', () => {
+    const user = [320, 375, 768];
```
bhokaremoin, line 5: as discussed found that test:coverage is not working correctly, where we able to figure and fix it.

src/snapshots.js
rahulXbrowserstack: for enableJS -> this is the implementation here - https://github.com/percy/percy-storybook/blob/833d07be059d51a950cea04683beab816596b7ae/src/snapshots.js#L212

Also tested with RSC -> false and enableJS true -> it was working as expected, same as selenium, checked for all widths and single width too, it was showing diff doms for all widths, which was same at selenium build

Also tested with RSC -> true/false and enable JS true/false -> the result was same as selenium checked for multiple widths and single width and for mobile browsers as well

bhokaremoin (changes requested):

test/utils.test.js
rahulXbrowserstack: yes, as said given 2 hrs there but couldn't found a fix for that, we would need extra time and effort for that, we can cover that in some seperate ticket, also i would suggest not to use jasmine , instead we should use jest , but we can consider this for future scope

src/snapshots.js
```diff
@@ -8,9 +8,52 @@ import {
...
+    const responsiveOptions = { ...options, responsiveSnapshotCapture: true, widths };
+    log.debug('captureDOM: Using responsive snapshot capture', { responsiveOptions });
+    return await captureResponsiveDOM(page, responsiveOptions, percy, log);
+  } else {
+    const eligibleWidths = {
+      config: percy.config.snapshot?.widths || []
+    };
+    widths = getWidthsForResponsiveCapture(
+      options.widths,
+      eligibleWidths
```
bhokaremoin, line 46: why are we passing getWidthsForResponsiveCapture when responsiveSnapshotCapture is false. we don't want to make any change in existing flow. no change in options that is send to page.snapshot and percy.snapshot

src/snapshots.js
rahulXbrowserstack: https://github.com/percy/percy-storybook/pull/1169#discussion_r2244506608

Please check here once, I replied

src/utils.js
```diff
@@ -417,3 +417,164 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+  try {
+    // Use Percy's CDP-based resize method
+    await page.resize({
+      width,
+      height,
+      deviceScaleFactor: 1,
+      mobile: false
+    });
+  } catch (e) {
+    log.debug('Resizing using CDP failed, falling back to page eval for width', width, e);
```
bhokaremoin, line 465: we don't want to raise error, if the resizing failed ?

src/utils.js
```diff
@@ -417,3 +417,164 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+        // Trigger resize events and force layout
+        window.dispatchEvent(new Event('resize'));
+        document.body.offsetHeight;
+      }, { width, height });
+    } catch (fallbackError) {
+      log.error('Fallback resize using page.eval failed', fallbackError);
+      throw new Error(`Failed to resize viewport using both CDP and page.eval: ${fallbackError.message}`);
+    }
+  }
+
```
bhokaremoin, line 479: also any reason we have 3 different try catch in single function, check if its possible to wrap in single try catch only

src/utils.js
```diff
@@ -417,3 +417,164 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+  try {
+    // Wait for resize to complete
+    await page.eval(({ resizeCount }) => {
+      return new Promise((resolve, reject) => {
+        const timeout = setTimeout(() => reject(new Error('Timeout')), 1000);
+        (function check() {
```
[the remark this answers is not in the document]
rahulXbrowserstack: resolved

src/utils.js
```diff
@@ -417,3 +417,164 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+          if (window.resizeCount === resizeCount) {
+            clearTimeout(timeout);
+            resolve();
+          }
+        })();
```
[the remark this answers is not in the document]
rahulXbrowserstack: change done

src/snapshots.js
rahulXbrowserstack: consider the scenario when user passes [a, b, c] widths at global level and [c, d] at snapshot level, then it has to be taken care, it has to give precendence to snapshot width in that case, this is the same behaviour happening at selenium side too

src/utils.js
rahulXbrowserstack: we are raising error, we have catch block at both places - its following the same practice as this - https://github.com/percy/percy-selenium-js/blob/acc54c599e81a62d1d55d2352508164a1db25ce6/index.js#L33

src/utils.js
rahulXbrowserstack: reason for having three separate try-catch blocks in changeViewportDimensionAndWait is to handle different failure scenarios independently:

- First try-catch: Attempts the primary resize method (CDP), and if it fails, falls back to a secondary method (page.eval). If both fail, it throws an error.

- Second try-catch (inside the first catch): Handles errors specifically from the fallback resize method, so you can log a different error and throw a more descriptive message.

- Third try-catch: Handles errors when waiting for the resize to complete, so a timeout or other error here doesn't prevent the function from continuing, but logs a debug message.

Also its the same way here as well ref: https://github.com/percy/percy-selenium-js/blob/acc54c599e81a62d1d55d2352508164a1db25ce6/index.js#L33

src/snapshots.js
bhokaremoin: how is this getting handled in current sdk ?

bhokaremoin: got it this is because we have introduced widths as an option on snapshot level.
then change the name of this function - `getWidthsForResponsiveCapture` to more generic one, since its getting used irrespective of responsiveSnapshotCapture

src/snapshots.js
bhokaremoin: got it.

bhokaremoin: `options.domSnapshot = previewResource.content;` for enable_javascript this dom is provided by storybook itself.

src/utils.js
bhokaremoin: got it, missed this, please add a small comment

test/utils.test.js
bhokaremoin: ok

test/utils.test.js
```diff
@@ -0,0 +1,147 @@
...
+      Promise.resolve({ domSnapshot: { html: '<html>resp</html>' } })
+    );
+  });
+
+  it('produces snapshots at each width in the list', async () => {
+    const options = { widths: [800, 1024] };
+    const result = await utils.captureResponsiveDOM(page, options, percy, log);
+
+    expect(Array.isArray(result)).toBe(true);
+    expect(result.length).toBe(2);
```
bhokaremoin, line 124: add check of widths

test/utils.test.js
```diff
@@ -0,0 +1,147 @@
...
+  });
+
+  it('ignores bad entries and trims spaces', () => {
+    expect(parse(' 375 , 768 , abc , 1280 ')).toEqual([375, 768, 1280]);
+  });
+
+  it('returns [] for an empty string', () => {
+    expect(parse('')).toEqual([]);
+  });
+});
```
bhokaremoin, line 147: test for `changeViewportDimensionAndWait` is missing please add

test/utils.test.js
bhokaremoin: will need to fix this, please create a ticket.

rahulXbrowserstack: sure

test/utils.test.js
rahulXbrowserstack: added

test/utils.test.js
rahulXbrowserstack: added that

utk787 (approved):

PR #1194, opened 2025-09-01
Capturing DOM at device height for mobile browsers + Fix for minHeight

rahulXbrowserstack:
**Change 1:**

- For mobile devices, earlier the DOM was being captured using the same height as desktop browsers, which didn’t reflect the actual view on mobiles.
- To fix this, we’ve added a new parameter height, which comes from the `device-details` API. This ensures the real device height is included when capturing.
- For example:
     - `iPhone 13` → width: `390px`, height: `663px`
     - `Samsung S22` → width: `360px`, height: `649px`
- Now, when a mobile device is selected, its DOM will be captured at the specific device height.
- If the same widths (390px or 360px) are used in desktop configs, then:
     - The default desktop height will be applied, OR
     - If `minHeight` is set and the `PERCY_RESPONSIVE_CAPTURE_MIN_HEIGHT` flag is enabled, then that `minHeight` will be used instead.

**Change 2:**

- This is a simple fix for the `PERCY_RESPONSIVE_CAPTURE_MIN_HEIGHT` flag. Earlier it wasn’t working correctly.
- Now:
     - If the flag is enabled and `minHeight` is set in the config → that value will be used.
     - If the flag is enabled but `minHeight` is not set → it will default to 1024.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -607,7 +607,7 @@ export async function captureResponsiveDOM(page, options, percy, log) {
   let defaultHeight = currentHeight;
   if (process.env.PERCY_RESPONSIVE_CAPTURE_MIN_HEIGHT) {
     const viewportOuterHeight = await page.eval(() => window.outerHeight);
-    const configMinHeight = percy?.config?.snapshot?.minHeight || 600;
+    const configMinHeight = percy?.config?.snapshot?.minHeight || 1024;
     defaultHeight = viewportOuterHeight - currentHeight + configMinHeight;
     log.debug(`Using custom minHeight for responsive capture: ${defaultHeight}`);
   }
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -530,11 +530,11 @@ describe('captureResponsiveDOM environment variables', () => {
     const options = { widths: [768] };
     await utils.captureResponsiveDOM(page, options, percy, log);
 
-    // Should calculate: outerHeight(800) - currentHeight(667) + fallback(600) = 733
-    expect(log.debug).toHaveBeenCalledWith('Using custom minHeight for responsive capture: 733');
+    // Should calculate: outerHeight(800) - currentHeight(667) + fallback(1024) = 1157
+    expect(log.debug).toHaveBeenCalledWith('Using custom minHeight for responsive capture: 1157');
     expect(page.resize).toHaveBeenCalledWith({
       width: 768,
-      height: 733,
+      height: 1157,
       deviceScaleFactor: 1,
       mobile: false
     });


src/utils.js
```diff
@@ -599,11 +655,15 @@ export async function captureResponsiveDOM(page, options, percy, log) {
...
-    log.debug(`Taking DOM snapshot at width=${width}`);
-    let domSnapshot = await captureSerializedDOM(page, options, log);
-    let snapshotWithWidth = { ...domSnapshot, width };
-    domSnapshots.push(snapshotWithWidth);
-    log.debug(`Snapshot captured for width=${width}`);
+    log.debug(`Taking DOM snapshot at width=${width}, height=${targetHeight}`);
+    const serializationOptions = {
+      ...options,
+      responsiveSnapshotCapture: true,
+      widths: widthHeightCombinations.map(item => item.width) // Convert to simple widths array
```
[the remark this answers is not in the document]
rahulXbrowserstack: can't be done this way

test/utils.test.js
```diff
@@ -101,6 +170,12 @@ describe('captureResponsiveDOM', () => {
     };
 
     percy = {
+      client: {
+        getDeviceDetails: jasmine.createSpy('getDeviceDetails').and.returnValue(Promise.resolve([
+          { width: 375, height: 812 },
+          { width: 414, height: 896 }
+        ]))
+      },
```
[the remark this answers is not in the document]
rahulXbrowserstack: done

src/snapshots.js
```diff
@@ -22,40 +22,16 @@ export async function captureDOM(page, options, percy, log) {
...
-    const singleDOMOptions = { ...options, widths };
-    log.debug('captureDOM: Using single snapshot capture', {
-      singleDOMOptions
-    });
-    return await captureSerializedDOM(page, singleDOMOptions, log);
+    log.debug('captureDOM: Using responsive snapshot capture', { options });
+    return await captureResponsiveDOM(page, options, percy, log);
   }
+
+  log.debug('captureDOM: Using single snapshot capture');
```
pankaj443, line 30: was it supposed to be in else block as previously?

src/utils.js
```diff
@@ -443,21 +443,64 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
+/**
+ * Build width-height combinations for responsive capture
+ */
+function buildWidthHeightCombinations(userPassedWidths, eligibleWidths, defaultHeight) {
+  const widthHeightMap = new Map();
+
+  // Add mobile devices with their heights
+  eligibleWidths.mobile?.forEach(device => {
+    if (device.width != null) {
+      widthHeightMap.set(device.width, device.height || defaultHeight);
```
pankaj443, line 467: is this device height post dpr calc?

src/utils.js
```diff
@@ -556,28 +596,44 @@ export async function captureResponsiveDOM(page, options, percy, log) {
...
+
+  // PERCY_RESPONSIVE_CAPTURE_MIN_HEIGHT: If set, overrides the minimum height for the viewport during capture.
+  let defaultHeight = currentHeight;
   if (process.env.PERCY_RESPONSIVE_CAPTURE_MIN_HEIGHT) {
-    height = await page.eval((configMinHeight) => {
-      return window.outerHeight - window.innerHeight + configMinHeight;
-    }, percy?.config?.snapshot?.minHeight || 600);
-    log.debug(`Using custom minHeight for responsive capture: ${height}`);
+    const viewportOuterHeight = await page.eval(() => window.outerHeight);
+    const configMinHeight = percy?.config?.snapshot?.minHeight || 600;
```
pankaj443: why default 600?

src/snapshots.js
rahulXbrowserstack: previously if block was handling RSC -> true and else block was handling when RSC -> false

Now we are handling RSC -> true and returning inside the if block so no need to else

prev ->
```
if() {
  code 1
}
else {
  code2
}
```

now ->
```
if() {
  code 1
  return
}
code 2
```

Also the other logics inside the if block when RSC -> true is moved to captureResponsiveDOM function

src/utils.js
rahulXbrowserstack: it will take care of the height after dpr by its own, we need to pass here like 663 and not 663 * dpr , same as width

Have ran multiple builds and it was working fine

src/utils.js
rahulXbrowserstack: It should be 1024, Thanks for pointing out, correcting it

pankaj443 (approved):
LGTM

prklm10 (approved):

PR #1196, opened 2025-09-04
Fix: Preserve story state & prevent blank/loader snapshots after reload

rahulXbrowserstack:
## Goal / Problem:

When the `PERCY_RESPONSIVE_CAPTURE_RELOAD_PAGE` flag was enabled, some snapshots were coming out **blank or stuck on loaders**.

This happened because:

*   We reloaded the page after resizing, but did not wait for the story to fully load again and for the loaders to disappear.
*   The URL from `window.location.href` didn't include story-specific params (like `globals`, `args`, `viewMode`), so the reloaded story lost its state.

### 1. Correct Reload URL

*   Instead of using `window.location.href`, we now build the full URL directly from `story.url`.
*   This ensures `viewMode=story` and all story params (`globals`, `args`) are included.

### 2. Re-apply Story State After Reload

```js
await page.eval(evalSetCurrentStory, story);
```

*   Guarantees that after reload, the story is restored with the correct state (e.g., dark mode, args, globals).
*   Also ensures loaders disappear and actual story content is present before capture.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -575,7 +575,7 @@ export async function captureSerializedDOM(page, options, log) {
 }
 
 // Capture responsive DOM snapshots across different widths
-export async function captureResponsiveDOM(page, options, percy, log) {
+export async function captureResponsiveDOM(page, options, percy, log, story) {
   const domSnapshots = [];
 
   // Get current viewport size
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -633,14 +633,21 @@ export async function captureResponsiveDOM(page, options, percy, log) {
...
+
       log.debug('Reloading page for responsive capture');
-      await page.goto(currentUrl, { forceReload: true });
+      await page.goto(reloadUrl, { forceReload: true });
 
-      // Re-inject PercyDOM if needed
-      await page.insertPercyDom();
+      // After reload, re-apply the story state (globals, args, etc.)
+      log.debug('Re-applying story state after reload');
+      await page.eval(evalSetCurrentStory, story);
```
bhokaremoin, line 650: 1. why is `await page.insertPercyDom();` removed ?
2. also didn't get why do we want to navigate to story.url and not reload on the existing url
shouldn't window.location.href be same as story.url
if there is some reasoning please mention it in comment or PR description for future understanding

rahulXbrowserstack: 1. There was no use of `await page.insertPercyDom();`, since its already been included into the page.snapshot() -> function, if you will see its implementation
2. window.location.href was not giving the entire url with globals and params in there if was just giving it with id
image

Like here in the image you can see we have id and globals etc, but that was giving just id

- story.url contains the complete story configuration (args, globals, queryParams)
- window.location.href may be missing story-specific parameters that were set during
- initial navigation, especially if the page was navigated to via programmatic means
- This ensures we reload to the exact same story state, not just the base URL

Its already included in DESC

bhokaremoin: got it

bhokaremoin (approved):
LGTM

test/snapshots.test.js
```diff
@@ -52,11 +54,19 @@ describe('captureDOM behavior', () => {
           allWidths = allWidths.concat(configWidths); // Fallback to config widths
         }
         const uniqueWidths = [...new Set(allWidths)].filter(w => w); // Remove duplicates
+
+        // Simulate the story state restoration behavior when PERCY_RESPONSIVE_CAPTURE_RELOAD_PAGE is set
+        if (process.env.PERCY_RESPONSIVE_CAPTURE_RELOAD_PAGE && story) {
+          log.debug('Reloading page for responsive capture');
+          log.debug('Re-applying story state after reload');
```
prklm10: two log lines are un-necessary.

rahulXbrowserstack: removed one

prklm10 (approved):

PR #1203, opened 2025-09-12
Await snapshot promise to avoid snapshots being taken of the wrong story

Thao-Tran:
My team is currently using `@percy/storybook` for taking snapshots of our component stories in our Linux CI environment and we've been running into flaky issues with percy taking a snapshot of the wrong story. I think it's due to the snapshot command not being awaited because I've been testing this fix in our CI runs and haven't run into the flaky snapshot issue yet.

### DIFF
--- a/src/snapshots.js
+++ b/src/snapshots.js
@@ -278,7 +278,7 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags 
               const options = yield* processStory(page, currentStory, previewResource, percy, flags, log);
 
               // Take the snapshot
-              percy.snapshot(options);
+              yield percy.snapshot(options);
 
               // Update previous story tracking
               previousStory = currentStory;


src/snapshots.js
```diff
@@ -278,7 +278,7 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
               const options = yield* processStory(page, currentStory, previewResource, percy, flags, log);
 
               // Take the snapshot
-              percy.snapshot(options);
+              yield percy.snapshot(options);
```
ninadbstack, line 281: capturing of story - `captureDOM` is called as part of `processStory` call above which is awaited - the resource discovery and posting it to percy.io is done as part of `percy.snapshot` call - which can be async. 

So it seems unlikely that the issue that you are facing is due to missing yield here as its intentional. But if this is in fact fixing your issue please raise a support ticket and we will try to find why.

PR #1206, opened 2025-09-29
MDX Snapshot Support For Percy Storybook

ayushatstack:
https://browserstack.atlassian.net/browse/PPLT-4544

1. Fetch doc ids from `window.__STORYBOOK_PREVIEW__.storyStoreValue.storyIndex.entries` and add them to the snapshot array

2. For docs we are using a fresh page to take snapshot because of context destruction

### DIFF
--- a/src/snapshots.js
+++ b/src/snapshots.js
@@ -166,11 +166,13 @@ function mapStorybookSnapshots(stories, { previewUrl, flags, config }) {
   if (!snapshots.length) throw new Error('No snapshots found');
 
   // remove filter options and generate story snapshot URLs
-  return snapshots.map(({ skip, include, exclude, ...story }) => {
+  return snapshots.map(({ skip, include, exclude, ...story }, i) => {
     let url = `${previewUrl}?id=${story.id}`;
     if (story.args) url += `&args=${buildStorybookArgsParam(story.args)}`;
     if (story.globals) url += `&globals=${buildStorybookArgsParam(story.globals)}`;
     for (let [k, v] of Object.entries(story.queryParams ?? {})) url += `&${k}=${v}`;
+    // Only add type if present in the original data
+    if (stories.data[i]?.type) story.type = stories.data[i].type;
     return Object.assign(story, { url });
   });
 }
@@ -185,7 +187,7 @@ function hasContaminatingState(story) {
 function needsFreshPage(previousStory) {
   // Only need fresh page if previous story had contaminating state
   // The current story will be loaded correctly regardless of globals/queryParams
-  return previousStory && hasContaminatingState(previousStory);
+  return previousStory && (hasContaminatingState(previousStory) || previousStory.type === 'docs');
 }
 
 // Process a single story and capture its DOM
@@ -232,7 +234,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags 
     // gather storybook data in parallel
     let [environmentInfo, stories] = yield* yieldAll([
       withPage(percy, aboutUrl, p => p.eval(evalStorybookEnvironmentInfo), undefined, { from: 'about url' }),
-      withPage(percy, previewUrl, p => p.eval(evalStorybookStorySnapshots), undefined, { from: 'preview url' })
+      withPage(
+        percy,
+        previewUrl,
+        p => p.eval(
+          evalStorybookStorySnapshots,
+          {
+            docCapture: process.env.PERCY_STORYBOOK_DOC_CAPTURE === 'true',
+            autodocCapture: process.env.PERCY_STORYBOOK_AUTODOC_CAPTURE === 'true'
+          }
+        ),
+        undefined,
+        { from: 'preview url' }
+      )
     ]);
 
     // map stories to snapshot options
--- a/src/utils.js
+++ b/src/utils.js
@@ -256,7 +256,7 @@ export function evalStorybookEnvironmentInfo({ waitForXPath }) {
 
 // Evaluate and return serialized Storybook stories to snapshot
 /* istanbul ignore next: no instrumenting injected code */
-export function evalStorybookStorySnapshots({ waitFor }) {
+export function evalStorybookStorySnapshots({ waitFor }, { docCapture = false, autodocCapture = false } = {}) {
   let serialize = (what, value, invalid) => {
     if (what === 'include' || what === 'exclude') {
       return [].concat(value).filter(Boolean).map(v => v.toString());
@@ -289,7 +289,22 @@ export function evalStorybookStorySnapshots({ waitFor }) {
 
     const storiesObj = await (window.__STORYBOOK_PREVIEW__?.extract?.());
     if (storiesObj && !Array.isArray(storiesObj)) {
-      return Object.values(storiesObj);
+      const stories = Object.values(storiesObj);
+
+      let docsEntries = [];
+      if (docCapture || autodocCapture) {
+        docsEntries = Object.values(window.__STORYBOOK_PREVIEW__.storyStoreValue.storyIndex.entries)
+          .filter(entry => {
+            if (entry.type !== 'docs') return false;
+            if (entry.tags && entry.tags.includes('autodocs')) {
+              return !!autodocCapture;
+            } else {
+              return !!docCapture;
+            }
+          });
+      }
+
+      return stories.concat(docsEntries);
     }
 
     await window.__STORYBOOK_STORY_STORE__?.extract?.();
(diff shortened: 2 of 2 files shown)


src/snapshots.js
```diff
@@ -61,17 +61,23 @@ function shouldSkipStory(name, options, config) {
 // Returns snapshot config options for a Storybook story merged with global Storybook
 // options. Validation error messages will be added to the provided validations set.
 function getSnapshotConfig(story, config, invalid) {
-  let { id, ...options } = PercyConfig.migrate(story, '/storybook');
+  // Remove type before validation
+  const { type, ...storyWithoutType } = story;
+  let { id, ...options } = PercyConfig.migrate(storyWithoutType, '/storybook');
```
bhokaremoin: why is this "type" removed. we need to validate type config also, if something other than doc or story is provided then we need to throw error. pass type and have the validation for it

src/snapshots.js
```diff
@@ -232,7 +238,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
       withPage(percy, aboutUrl, p => p.eval(evalStorybookEnvironmentInfo), undefined, { from: 'about url' }),
-      withPage(percy, previewUrl, p => p.eval(evalStorybookStorySnapshots), undefined, { from: 'preview url' })
+      withPage(
+        percy,
+        previewUrl,
+        p => p.eval(
+          evalStorybookStorySnapshots,
+          {
+            docCapture: process.env.PERCY_STORYBOOK_DOC_CAPTURE === 'true',
+            autodocCapture: process.env.PERCY_STORYBOOK_AUTODOC_CAPTURE === 'true'
```
bhokaremoin: store this check in a flag and use that flag value

src/utils.js
```diff
@@ -256,7 +256,7 @@ export function evalStorybookEnvironmentInfo({ waitForXPath }) {
 
 // Evaluate and return serialized Storybook stories to snapshot
 /* istanbul ignore next: no instrumenting injected code */
-export function evalStorybookStorySnapshots({ waitFor }) {
+export function evalStorybookStorySnapshots({ waitFor }, { docCapture = false, autodocCapture = false } = {}) {
```
bhokaremoin, line 259: why not - 
`export function evalStorybookStorySnapshots({ waitFor, docCapture = false, autodocCapture = false } = {})`

src/utils.js
```diff
@@ -301,9 +316,10 @@ export function evalStorybookStorySnapshots({ waitFor }) {
     let invalid = new Map();
 
     let data = stories.map(story => serialize('snapshot', {
-      name: `${story.kind}: ${story.name}`,
+      name: story.kind ? `${story.kind}: ${story.name}` : `${story.title}: ${story.name}`,
```
bhokaremoin, line 319: why this change, this will name the snapshot name of user not using docs feature, this will affect the baselines

test/utils.test.js
```diff
@@ -746,3 +746,78 @@ describe('captureResponsiveDOM story state restoration', () => {
...
+    delete global.window.__STORYBOOK_PREVIEW__.storyStoreValue.storyIndex.entries;
+    const { data } = await utils.evalStorybookStorySnapshots(
+      { waitFor }, { docCapture: true, autodocCapture: true }
+    );
+    const ids = data.map(s => s.id);
+    expect(ids).toContain('story--one');
+    expect(ids).not.toContain('docs--one');
+    expect(ids).not.toContain('autodoc--one');
+    expect(data.length).toBe(1);
+  });
```
bhokaremoin, line 822: rspec for validation is if - process.env.PERCY_STORYBOOK_DOC_CAPTURE and process.env.PERCY_STORYBOOK_AUTODOC_CAPTURE is working

test/utils.test.js
```diff
@@ -746,3 +746,78 @@ describe('captureResponsiveDOM story state restoration', () => {
...
+});
```
bhokaremoin, line 823: rspecs for changes made in snapshots.js

src/snapshots.js
```diff
@@ -232,7 +238,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
+          }
+        ),
```
bhokaremoin, line 246: since this is a open source repo, check if any readme changes are required

bhokaremoin (changes requested):

src/utils.js
ayushatstack: the data format of doc entries is a bit different they don't contain kind
For normal stories we get the old behaviour - ${story.kind}: ${story.name}`  same as before
for doc entries - `${story.title}: ${story.name}` this has been discussed with product

test/utils.test.js
ayushatstack: `includes story, docs, and autodoc when both flags are true` it's this one

src/utils.js
ayushatstack: just to group filtering option in a single object

src/snapshots.js
ayushatstack: done

test/utils.test.js
bhokaremoin: no I mean, pass this value in env for test and check if that are getting passed correctly, - "includes story, docs, and autodoc when both flags are true" this is functional test

src/utils.js
bhokaremoin: ok, just make sure the product callout and discussion is present in DB questionnaires

ayushatstack: prev discussion - https://browserstack.atlassian.net/browse/PER-5667?focusedCommentId=1568216&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&sourceType=mention#comment-1568216

src/snapshots.js
ayushatstack: we have Supported environment parameters 
in the online doc - https://www.browserstack.com/docs/percy/integrate/storybook but not in the readme. Shouldn't we add the mdx env variable to the env when release stable version though? Will add it then

test/utils.test.js
ayushatstack: done in snapshots.test.js

test/utils.test.js
ayushatstack: done in snapshots.test.js

test/snapshots.test.js
```diff
@@ -232,3 +232,61 @@ describe('captureDOM behavior', () => {
...
+      eval: jasmine.createSpy('eval').and.returnValue(
+        Promise.resolve({ invalid: [], data: [{ id: 'story--one', name: 'Story One' }] })
+      ),
+      snapshot: jasmine.createSpy('snapshot'),
+      resize: jasmine.createSpy('resize'),
+      goto: jasmine.createSpy('goto'),
+      insertPercyDom: jasmine.createSpy('insertPercyDom')
+    };
+    withPage = (_percy, url, cb) => cb(page);
+    takeStorybookSnapshots = function*(percy, callback, { baseUrl, flags }) {
```
ayushatstack, line 250: created a stub to mock the structure of requests. evalStorybookStorySnapshots is called via page.eval by wrapping it in withPercyStorybookHelpers and executing it in the page. This execution is difficult to mock and as the function execute in page.

the strurture here mimics the same request pattern

src/snapshots.js
bhokaremoin: got it

src/utils.js
bhokaremoin: update the check - have a check on story.type, as of now story.kind is not present for docs so this condition works. but storybook can change this in future, so lets keep the check on type only.

src/utils.js
```diff
@@ -335,14 +351,17 @@ export function evalSetCurrentStory({ waitFor }, story) {
...
         // This helps ensure that any post-render loader state changes have time to occur
         setTimeout(() => {
           // Wait for any loaders to disappear before resolving
           waitForLoadersToDisappear().then(resolve).catch(reject);
         }, 100);
-      });
+      };
+
+      channel.on('storyRendered', handleRendered);
+      channel.on('docsRendered', handleRendered);
```
bhokaremoin, line 364: rspecs for this change are missing.
Currently there is some issue with test coverage, but we need to make sure there are atleast unit testing present for this.

src/utils.js
ayushatstack: type also only exists for docs that why we do type: story.type ? story.type : 'story'.  title and kind also contain the same value - Example/Button1

bhokaremoin: how is kind related to type ?
story have kind and docs does not correct.
I am saying then why not have check on story.type itself 
story.type == docs -> `${story.title}: ${story.name}`
else -> `${story.kind}: ${story.name}`

bhokaremoin: this will keep snapshot name consistent even if later kind field is added to docs by storybook. we don't want upgrading storybook version to affect the snapshot names of customer

bs-shobhitkumar (approved):
lgtm, check the specs comments from @bhokaremoin

src/utils.js
ayushatstack: sorted over call

src/utils.js
ayushatstack: done

bhokaremoin (approved):

PR #1289, opened 2026-04-23
fix(build): guard against dev-React leakage into manager bundle

prklm10:
## Summary

Resolves the React 19 manager crash in `@percy/storybook` on Storybook 10 (the `Cannot read properties of undefined (reading 'recentlyCreatedOwnerStacks')` crash that wiped the manager UI).

## Root cause (empirically confirmed)

Storybook 10's manager runs **production-minified React 19** (`__REACT__` global). Its manager-builder uses esbuild's globals-externals plugin which destructures named React exports from `__REACT__` into a local `react_exports` object — and the destructure **does not include** `__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE`. In `storybook dev`, the manager-builder also pulls in `react-jsx-runtime.development.js`, whose top-level reads:

```js
var ReactSharedInternals = React.__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE;
```

Because the destructure stripped that property and prod React 19 doesn't expose it anyway, `ReactSharedInternals` is undefined. Every `_jsx()` call from our addon (e.g. `PanelTitle`, `renderLabel`) crashes on `ReactSharedInternals.recentlyCreatedOwnerStacks++`. `ManagerErrorBoundary` catches → blank sidebar.

Patching `__REACT__` at runtime is futile: by the time our shim runs, the destructure has already produced `react_exports` without the stripped property.

### 1. `vite.config.mjs` — switch to classic JSX
```diff
 react({
-  jsxRuntime: 'automatic',
+  jsxRuntime: 'classic',
```
Our compiled bundle now emits `React.createElement(...)` calls instead of `_jsx(...)`. `createElement` uses only public React APIs that ARE in esbuild's destructure, so the prod React in the manager works fine. The dev jsx-runtime that Storybook bundles becomes **dead code** — present but never called. Bundle size delta: +6 KB (negligible).

### 2. `src/react19Shim.js` — defense-in-depth for non-classic deps
Rewrote to **create** `__CLIENT_INTERNALS_*` with stub dev fields when prod React 19 strips them. This is the original brainstorm direction; it doesn't fix the crash by itself (destructure mismatch — see Root cause), but it covers any bundled dep that compiles with `automatic` JSX and bypasses our classic-JSX choice. Combined with #1, both code paths are safe.

### 3. `scripts/assert-prod-bundle.mjs` — allow-list update
Added `recentlyCreatedOwnerStacks` to the shim's self-invalidating allow-list (the rewritten shim legitimately contains this token in its create-if-missing block).

Build-time guard from earlier in this PR remains in place.

## Verification

Fixture at `~/percyApps/percy-storybook-fixtures/sb10-vite/` — clean `yarn pack` install on Node 22.20.0, Storybook 10.3.5, React 19.2.5, `@storybook/react-vite@10.3.5`.

- [x] `yarn build` — `✓ dist/ is production-clean (6 files scanned)`.
- [x] `yarn lint` — pass.
- [x] Self-invalidation: `mv src/react19Shim.js /tmp/ && node scripts/assert-prod-bundle.mjs` exits 1 with allow-list-deletion message; restoring → exits 0.
- [x] **Live `storybook dev` + headless Chromium (Playwright):** sidebar renders 2 items; **zero `recentlyCreatedOwnerStacks` / `ReactSharedInternals` errors** in `pageerror` or `console.error`. Manager boots clean.
- [x] Bundle inspection: live `manager-bundle.js` has 1721 `createElement` call sites, near-zero `_jsx(` callers (bundled deps still produce some); the dev jsx-runtime tokens are present but uncalled from our code.

## Post-Deploy Monitoring & Validation

- **What to monitor**
  - Logs: N/A. Build/runtime fix only.
  - Metrics: GitHub Actions PR `build` step — watch for >+5% time. Local `yarn build` on this PR runs ~46s vs ~20s baseline; the increase is from Vite re-tree-shaking with classic JSX. Tolerable; expected to amortize on warm caches.
- **Validation checks**
  - On every fresh install in any consuming Storybook 10 + React 19 project:
    ```bash
    npm install @percy/storybook && npx storybook dev
    # Manager loads, sidebar renders, no recentlyCreatedOwnerStacks errors in console.
    ```
- **Expected healthy behavior**
  - `yarn build` → `✓ dist/ is production-clean`.
  - Consumer's manager DevTools console: zero `recentlyCreatedOwnerStacks` / `ReactSharedInternals` errors.
  - Sidebar populated, Percy panel tab reachable.
- **Failure signal(s) / rollback trigger**
  - Bundle grep failure → investigate which dep introduced the regression; do not bypass.
  - User reports of `recentlyCreatedOwnerStacks` post-merge → revert this PR; reopen with a refined approach (likely: also patch `react_exports` via a Vite plugin that intercepts esbuild output).
- **Validation window & owner**
  - 14 days post-merge on `master`. Owner: @prklm10.

## Files changed

- `vite.config.mjs` — `jsxRuntime: 'classic'`
- `src/react19Shim.js` — rewrite to create `__CLIENT_INTERNALS_*` when missing; backfill stub fields
- `scripts/assert-prod-bundle.mjs` — extend allow-list with `recentlyCreatedOwnerStacks`
- (earlier in this PR) `scripts/assert-prod-bundle.mjs` (new), `package.json` build hook

## Follow-ups (separate)

- Phase 3 — CI matrix workflow with Playwright headless boot test running against `storybook dev` (matches user-facing crash repro).
- Phase 4 — `10.0.1-beta.1` release, narrow `peerDependencies.storybook` to `^9 || ^10`, README compat matrix.
- Phase 5 — Nightly canary against `react@next` + `storybook@next`.
- Investigate if @storybook/react-vite's manager-builder dev-jsx injection can be configured/skipped upstream.

---

[Compound Engineered](https://github.com/EveryInc/compound-engineering-plugin) 🤖 Generated with [Claude Code](https://claude.com/claude-code)


prklm10:
## Update: pivoted to bundling prod `react/jsx-runtime` (commit 1ec5bd2)

The earlier classic-JSX-plus-shim approach in this PR didn't actually fix the crash in the customer demo (`~/Downloads/percy-storybook-demo`, React 19 + Storybook 10.3.5). Re-investigated and found a different root cause than originally documented above.

### What I missed before

Storybook 10.3.3+ bundles **React 18** in its manager (`__REACT__` global), not React 19 — verified by reading `node_modules/storybook/dist/manager/globals-runtime.js:13401`, which shows the bundled prod jsx-runtime literally calls `f.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED.ReactCurrentOwner` (the React 18 internals key, not React 19's `__CLIENT_INTERNALS_*`).

### Actual chain

1. Customer has React 19 in their project's `node_modules`.
2. Storybook 10's manager-bundler resolves our addon's externalized `import "react/jsx-runtime"` against the customer's `node_modules` and concatenates **React 19's dev jsx-runtime** into our `manager-bundle.js`.
3. That dev runtime evaluates at module-init: `var ReactSharedInternals = React.__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE` — undefined on Storybook's globalized React 18.
4. First `_jsx()` call → `Cannot read properties of undefined (reading 'recentlyCreatedOwnerStacks')`.

### Why the shim couldn't win

ESM spec runs **all** `import` statements before any module body code. Verified directly in `dist/chunks/manager--*.js` — the rebundled output had `import { jsx } from "react/jsx-runtime"` hoisted to the top, four lines above where the shim body could install stubs. No bundling phase, banner, or `intro` can change ESM evaluation order for hoisted imports.

### New fix (commit 1ec5bd2)

Stop externalizing `react/jsx-runtime` in `vite.config.mjs`. Vite bundles **React 18's prod jsx-runtime** (~1KB minified) from our own `node_modules/react@18.3.1` into our chunks. It accesses `__SECRET_INTERNALS_*.ReactCurrentOwner` — which IS present on Storybook's globalized React 18 — so the panel renders.

Storybook's manager-bundler never sees a `react/jsx-runtime` import in our output, so it can't inject the customer's React 19 dev runtime. Customer React version becomes irrelevant for our manager bundle.

### Cascading cleanup in same commit

- Deleted `src/react19Shim.js` (now dead code; the file the shim guarded against is no longer imported).
- Pruned `scripts/assert-prod-bundle.mjs`: dropped `ALLOWED_CONTAINERS`, the existence-check loop, and `allowedTokensFor()`. Removed `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` from `FORBIDDEN` (now legitimately accessed by the bundled prod jsx-runtime).
- `manager.jsx` no longer imports the shim.
- Build script now does `rm -rf dist` upfront so deleted-source-file orphans (like the shim) can't persist.

### Verified

- `yarn build` → `✓ dist/ is production-clean (6 files scanned)`.
- `grep -c '"react/jsx-runtime"' dist/manager.js dist/chunks/*.js` → 0 (no externalized import remains).
- Bundled chunk contains React 18 prod jsx-runtime markers (`__SECRET_INTERNALS_*.ReactCurrentOwner`, `Symbol.for("react.element")`, `Symbol.for("react.fragment")`) — confirms inlining worked.
- Repacked tarball contains zero `react19Shim` references (54 files, neither `src/` nor `dist/` copy present).
- `yarn lint` passes.
- **End-to-end:** opened the Percy panel in `~/Downloads/percy-storybook-demo` after fresh tarball install — panel renders cleanly, no `recentlyCreatedOwnerStacks` crash. Confirmed by user.

### Caveats

- `yarn test` reports 45 failures + 1 error, but they pre-exist on the branch baseline (verified via `git stash` + re-run). The `pretest` hook's `storybook build` step exits 1 with `TypeError: Cannot read properties of undefined (reading 'kill')`. Tracked separately; not blocking this fix.
- A separate `"percy-storybook was loaded twice"` warning is visible in the demo console — likely caused by demo's `.storybook/main.js` listing the addon AND the postinstall auto-registration adding it. Out of scope here; track as follow-up.
- Not validated against SB8/SB9 fixtures (none locally available). Risk is low because the change only narrows externals.

### Files changed (final commit)

| File | Change |
|---|---|
| `vite.config.mjs` | Removed `react/jsx-runtime` from externals; added explanatory comment |
| `manager.jsx` | Removed shim import (line 8) |
| `src/react19Shim.js` | **Deleted** |
| `scripts/assert-prod-bundle.mjs` | Dropped allow-list machinery; tightened `FORBIDDEN`; updated stale plan reference |
| `package.json` | Build script now `rm -rf dist` upfront |

Net: **+20 / −111 lines.** PR description above is stale relative to commit 1ec5bd2 — the **classic JSX choice in `vite.config.mjs` is preserved**, but the rest of that flow is replaced. Will let @prklm10 decide whether to rewrite the PR description or leave the comment as the latest-state record.

PR #1292, opened 2026-04-30
PPLT-5302: smartsnap changes in storybook

RaghavsBrowserStack:
## What & Why

Wires the new **SmartSnap** feature into the Storybook SDK. SmartSnap reduces visual-test cost by snapshotting only the stories whose dependency graph changed relative to a baseline, instead of the full story set on every build.

The SmartSnap engine itself lives in `@percy/cli-command` (`applySmartSnap`, `SmartSnapBailError`). This PR is the **Storybook-side integration**: surfacing config/flags, extracting each story's source file (`importPath`) so the engine can map a snapshot back to the file that defines it, and invoking the filter at the right point in the snapshot pass.

Jira: PPLT-5302

## Changes

- **`src/common.js` / `src/config.js`** — Adds 7 `smart-snap-*` CLI flags and the matching `storybook.smartSnap.*` config schema:
  - `enabled` — turn SmartSnap on (default `false`)
  - `baseline` — git commit/branch to diff against
  - `untraced` — glob patterns excluded from dependency tracing
  - `trace` — verbose trace output for debugging
  - `bailOnChanges` — glob patterns that force a full build when changed
  - `statsFile` — name of the stats JSON inside the build dir (default `enriched-stats.json`)
  - `failBuildOnFailure` — fail the build on a SmartSnap error instead of falling back to a full snapshot set (default `false`)
- **`src/utils.js`** — Stamps each story with its `importPath` (preferring the v7+ `storyIndex.entries` map, falling back to `parameters.fileName` on older Storybook), carries it into the serialized snapshot `data`, and emits a `__smartsnapDiagnostics` blob (story counts, source, sample entries) for debug logging.
- **`src/snapshots.js`** — Preserves `importPath` through `/storybook` schema validation (which is `unevaluatedProperties: false`) and the config merge, strips it before DOM capture so it never leaks into a snapshot, and calls `applySmartSnap` after stories are mapped. Errors fall back to the full snapshot set; a `SmartSnapBailError` is logged at info level, and `failBuildOnFailure` re-throws.
- **`src/storybook.js`** — Threads `buildDir` (`args.serve`) into `takeStorybookSnapshots` so SmartSnap can locate the stats file.

## Behavior

- **Off by default** — no behavior change unless `smartSnap.enabled` (or `--smart-snap`) is set.
- **Fail-safe** — any SmartSnap error falls back to a normal full snapshot run, unless `failBuildOnFailure` is enabled.
- `importPath` is internal plumbing only; it is removed before capture and does not appear in the final snapshot.

## Dependencies / merge gating

- Requires a `@percy/cli` release that exports `applySmartSnap` and `SmartSnapBailError` from `@percy/cli-command`. The pinned `^1.31.10` does not yet include these — bump the dependency once that release is published.

## Testing

- [ ] Add coverage for `importPath` extraction (both Storybook-version paths), strip-before-capture, and the fallback vs. `failBuildOnFailure` branches — currently untested.

> Note: local `link:` dependency overrides and a stray `pnpm-lock.yaml` are kept out of this PR (the repo uses `yarn.lock`).


src/smartsnap.js
```diff
@@ -0,0 +1,283 @@
...
+
+  // Treat statsFile as a flat filename anchored inside the build directory.
+  // path.basename() strips any traversal segments so the resolved path can't
+  // escape buildDir even if the config is hostile.
+  const statsName = path.basename(statsFile || 'enriched-stats.json');
+  if (!/^[\w.-]+\.json$/i.test(statsName)) {
+    log.warn(`SmartSnap: invalid statsFile "${statsName}" — must be a .json filename; running full snapshot set`);
+    return snapshots;
+  }
+  const resolvedStatsPath = path.join(path.resolve(buildDir), statsName);
```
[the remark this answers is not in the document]
RaghavsBrowserStack: it is sanitized, the statsName file is validated to indeed be a json file as well as a file name and not a path + json.parse cannot be used for code execution. All results of this file read are treated as POD

RaghavsBrowserStack:
# Claude Code PR Review

**PR:** https://github.com/percy/percy-storybook/pull/1292  •  **Head:** 01b9dac  •  **Reviewers:** stack:percy-cli-code-reviewer

## Summary
Wires up "SmartSnap" for `percy-storybook`: adds 7 `smart-snap-*` CLI flags and a matching `storybook.smartSnap.*` config schema, stamps each story with its source `importPath` (v7+ `entries` map, falling back to `parameters.fileName`), threads `buildDir` (`args.serve`) down to `takeStorybookSnapshots`, and calls `applySmartSnap` (from `@percy/cli-command`) to filter the snapshot set before the capture loop, falling back to the full set on error. A `__smartsnapDiagnostics` blob is also collected during browser-side extraction.

> Note: per the PR author, the `applySmartSnap` / `SmartSnapBailError` exports from `@percy/cli-command` are not yet published — that is tracked separately and was intentionally excluded from this review.

## Review Table

| Priority | Category | Check | Status | Notes |
|----------|----------|-------|--------|-------|
| High | Security | No hardcoded secrets or credentials | Pass | No secrets/tokens in diff |
| High | Security | Authentication/authorization checks present | N/A | No auth surface touched |
| High | Security | Input validation and sanitization | Pass | Config goes through PercyConfig schema; see schema-laxity note below |
| High | Security | No IDOR — resource ownership validated | N/A | No resource access |
| High | Security | No SQL injection (parameterized queries) | N/A | No DB |
| High | Correctness | Logic is correct, handles edge cases | Fail | `buildDir` is `undefined` for URL-based runs (`storybook.js:44`); diagnostics computed unconditionally |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Fail | `SmartSnapBailError` + `failBuildOnFailure` contradictory: an intentional bail aborts the build (`snapshots.js:336-341`) |
| High | Correctness | No race conditions or concurrency issues | Pass | Sequential generator flow unchanged |
| Medium | Testing | New code has corresponding tests | Fail | Zero tests added for any new logic |
| Medium | Testing | Error paths and edge cases tested | Fail | None of the 3 catch branches / importPath paths covered |
| Medium | Testing | Existing tests still pass (no regressions) | N/A | Not run; no existing specs modified, but hot path (`mapStorybookSnapshots`) is touched |
| Medium | Performance | No N+1 queries or unbounded data fetching | Pass | Diagnostics blob computed per extraction — minor, see note |
| Medium | Performance | Long-running tasks use background jobs | N/A | N/A |
| Medium | Quality | Follows existing codebase patterns | Fail | `smartSnap` schema omits `unevaluatedProperties: false` used by every sibling object |
| Medium | Quality | Changes are focused (single concern) | Pass | Scoped to SmartSnap wiring |
| Low | Quality | Meaningful names, no dead code | Pass | `__smartsnapDiagnostics` is debug scaffolding left always-on — see note |
| Low | Quality | Comments explain why, not what | Pass | Comments are why-oriented and helpful |
| Low | Quality | No unnecessary dependencies added | Pass | No new deps |

## Findings

- **File:** `src/snapshots.js:336-341`
- **Severity:** Medium
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** `SmartSnapBailError` semantics contradict `failBuildOnFailure`. A bail is an *intentional* fall-back to the full snapshot set, not a failure — but when `failBuildOnFailure: true`, the catch logs the bail at `info` and then re-throws, aborting the build. A user enabling `failBuildOnFailure` to guard against unexpected SmartSnap errors (e.g. missing stats file) would also see normal bail conditions (e.g. a global CSS change) abort the build. The bail `info` log also doesn't state that the run continues with the full set.
- **Suggestion:** Only honour `failBuildOnFailure` for non-bail errors:
  ```js
  } catch (e) {
    if (e instanceof SmartSnapBailError) {
      log.info(`SmartSnap bailed (${e.message}); running full snapshot set`);
    } else {
      log.warn(`SmartSnap failed (${e.message}); running full snapshot set`);
      if (storybookConfig.smartSnap.failBuildOnFailure) throw e;
    }
  }
  ```

- **File:** `src/config.js:210-241`
- **Severity:** Medium
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** The `smartSnap` object schema omits `unevaluatedProperties: false`, which every other nested config object in this file (`additionalSnapshots`, `docsRule`, root snapshot object at lines 71/120/146/172) uses. A typo like `storybook.smartSnap.enabed: true` passes validation silently and SmartSnap stays disabled with no warning. (Verified against `src/config.js`.)
- **Suggestion:** Add `unevaluatedProperties: false` to the `smartSnap` schema object.

- **File:** `src/utils.js:344-370` (and host log `src/snapshots.js:315-317`)
- **Severity:** Medium
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** The `__smartsnapDiagnostics` blob is built on *every* `evalStorybookStorySnapshots` call — including production runs with SmartSnap disabled — and the host-side `log.debug` is guarded only by `stories?.diagnostics`, which is always populated. This is debug scaffolding running unconditionally and serialized across CDP on every extraction.
- **Suggestion:** Gate diagnostics computation behind a `withDiagnostics`/SmartSnap-enabled flag (like `docCapture`/`autodocCapture`) so it only runs when used.

- **File:** `src/storybook.js:44`
- **Severity:** Low
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** `buildDir: args.serve` is `undefined` when the user passes a live `--url` instead of a `--serve` directory, so SmartSnap against a running Storybook server has no build directory to locate `enriched-stats.json`, likely yielding an unhelpful downstream error. No guard or warning at the callsite.
- **Suggestion:** Warn when `storybookConfig?.smartSnap?.enabled && !buildDir` that SmartSnap requires a static build directory.

- **File:** `src/utils.js:362`
- **Severity:** Low
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** `stories.__smartsnapDiagnostics = {...}` attaches a named property to an Array. It works in-browser and is read correctly in `.then()`, so no runtime bug, but it is an unconventional shape that will confuse maintainers.
- **Suggestion:** Return `{ stories, diagnostics }` from the `waitFor` callback and destructure in `.then()`.

- **File:** (whole branch) — Testing
- **Severity:** Medium
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** No tests were added. Uncovered: the three `applySmartSnap` catch branches (bail / generic-fallback / `failBuildOnFailure` rethrow), `importPath` resolution (v7+ `entries`, `parameters.fileName` fallback, neither present), `importPath` survival through `getSnapshotConfig` → `mapStorybookSnapshots` → `processStory`, `buildDir` threading, and the diagnostics shape. This modifies the hot path that runs for every build.
- **Suggestion:** Add unit specs for the `applySmartSnap` block (all three catch branches) and for `importPath` extraction in `evalStorybookStorySnapshots`, at minimum.

---

**Verdict: FAIL** — directionally sound (importPath plumbing through the serializer is correct, fall-back-on-error is the right model, flag/schema structure follows convention), but request changes before merge: fix the `SmartSnapBailError`/`failBuildOnFailure` contradiction, add `unevaluatedProperties: false` to the `smartSnap` schema, gate the always-on diagnostics, and add test coverage for the new hot-path logic.

src/snapshots.js
```diff
@@ -322,6 +322,10 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
       )
     ]);
 
+    if (stories?.diagnostics) {
```
this-is-shivamsingh, line 326: What does this do ?

src/utils.js
```diff
@@ -312,14 +312,62 @@ export function evalStorybookStorySnapshots({ waitFor }, { docCapture = false, a
...
     if (storiesObj && !Array.isArray(storiesObj)) {
-      const stories = Object.values(storiesObj);
-      return stories.concat(docsEntries);
+      source = '__STORYBOOK_PREVIEW__.extract';
+      stories = stampImportPath(Object.values(storiesObj))
+        .concat(stampImportPath(docsEntries));
+    } else {
+      source = '__STORYBOOK_STORY_STORE__.raw';
+      await window.__STORYBOOK_STORY_STORE__?.extract?.();
+      stories = stampImportPath(window.__STORYBOOK_STORY_STORE__.raw());
```
this-is-shivamsingh, line 342: does it run when smart snap is disabled
- Question, do we need to run stampImportpath when there is smart snap disabled
- Or, this flow never reaches when smart snap is disabled

src/snapshots.js
RaghavsBrowserStack: The smartsnap flow needs the file path of each story. If we fail to extract this information the flow fails. Although storybook does export file path since v7, in case of some future regression if the path extraction logic fails it can be used to debug why.

src/utils.js
RaghavsBrowserStack: yeah, it runs when smartsnap is disabled, but it doesn't have any measurable performance impact. want me to gate it?

this-is-shivamsingh: This is operation is of no use, when disabled then Better to gate it

RaghavsBrowserStack: done

this-is-shivamsingh (approved):

RaghavsBrowserStack:
# Claude Code PR Review

**PR:** https://github.com/percy/percy-storybook/pull/1292  •  **Head:** 87c62bf  •  **Reviewers:** stack:percy-cli-code-reviewer, stack:percy-cli-code-review

## Summary
Adds SmartSnap support to the percy-storybook SDK: 7 new `storybook.smartSnap.*` CLI flags + config schema, browser-side `importPath` extraction and diagnostics (both gated behind a `smartSnap` flag), and wiring into `takeStorybookSnapshots` via `applySmartSnap` / `SmartSnapBailError` imported from `@percy/cli-command`. Bumps `@percy/cli-command` and `@percy/config` from `^1.31.10` to `1.32.0` (lockstep `yarn.lock` cascade).

> **Note on the prior `failBuildOnFailure` finding:** A previous review flagged that the catch block re-throws even a `SmartSnapBailError` when `failBuildOnFailure` is true, calling it a contradiction. This is **intentional and correct by design** and was explicitly excluded from this review. SmartSnap exists to save cost by reducing snapshot count; if it fails to reduce the count for *any* reason — including a bail — a customer who opted into `failBuildOnFailure` wants the build to hard-fail rather than silently run the expensive full snapshot set. Both reviewers confirmed they are not flagging this.

## Review Table

| Priority | Category | Check | Status | Notes |
|----------|----------|-------|--------|-------|
| High | Security | No hardcoded secrets or credentials | Pass | No `PERCY_TOKEN` / secrets in new code paths or fixtures. |
| High | Security | Authentication/authorization checks present | N/A | No auth surface in this change. |
| High | Security | Input validation and sanitization | Pass | `statsFile` path traversal mitigated in `@percy/cli-command` (`path.basename` + `/^[\w.-]+\.json$/i`); config value passed through cleanly. |
| High | Security | No IDOR — resource ownership validated | N/A | No multi-tenant resource access. |
| High | Security | No SQL injection (parameterized queries) | N/A | No database / SQL. |
| High | Correctness | Logic is correct, handles edge cases | Pass | `entries` is in scope (`utils.js:303`); `importPath` correctly stamped, merged, and stripped before DOM submission; `yield applySmartSnap` awaits correctly in the async generator; array-property diagnostics survive `.map`. |
| High | Correctness | Error handling is explicit, no swallowed exceptions | Pass | Bail vs warn branches explicit; `failBuildOnFailure` re-throw intentional. |
| High | Correctness | No race conditions or concurrency issues | Pass | No new shared mutable state across async boundaries. |
| Medium | Testing | New code has corresponding tests | Fail | `utils.js` importPath/diagnostics well covered (7 new specs), but the `applySmartSnap` try/catch wiring in `snapshots.js:341-352` has **zero** test coverage. |
| Medium | Testing | Error paths and edge cases tested | Fail | Bail path, `failBuildOnFailure` re-throw, generic-error fallback, and `buildDir === undefined` (URL/`start` mode) bail are untested. |
| Medium | Testing | Existing tests still pass (no regressions) | Pass | No existing specs modified; importPath stripped before validation/submission. |
| Medium | Performance | No N+1 queries or unbounded data fetching | Pass | importPath stamping is a single pass; diagnostics sample only the first entry/story. |
| Medium | Performance | Long-running tasks use background jobs | N/A | Not applicable. |
| Medium | Quality | Follows existing codebase patterns | Pass | Flag `percyrc`/`type`/`multiple`/`default` and `unevaluatedProperties:false` match surrounding conventions. |
| Medium | Quality | Changes are focused (single concern) | Pass | Scoped to SmartSnap enablement + dep bump. |
| Low | Quality | Meaningful names, no dead code | Pass | Clear names; gated helpers (`resolveImportPath`/`stampImportPath`) are tidy. |
| Low | Quality | Comments explain why, not what | Pass | Gating comment in `utils.js` explains the SmartSnap-only rationale. |
| Low | Quality | No unnecessary dependencies added | Pass | New transitive deps (`stream-json`, optional `snyk-nodejs-lockfile-parser`, grpc) come from the upstream `@percy/*` 1.32.0 bump, not added here. |

## Findings

- **File:** `src/snapshots.js:341-352`
- **Severity:** Medium
- **Reviewer:** stack:percy-cli-code-reviewer, stack:percy-cli-code-review (both)
- **Issue:** The new `if (storybookConfig?.smartSnap?.enabled)` try/catch — the load-bearing integration of this feature — has no test coverage. The bail-vs-warn branch, the intentional `failBuildOnFailure` re-throw, and the `yield applySmartSnap(...)` plumbing are all unexercised.
- **Suggestion:** Add specs in `test/snapshots.test.js`: (a) `SmartSnapBailError` → `log.info`, no throw when `failBuildOnFailure` is false; (b) generic error → `log.warn(... running full snapshot set)`, no throw; (c) both error types re-thrown when `failBuildOnFailure` is true; (d) `buildDir` undefined (URL/`start` mode) bails gracefully.

- **File:** `src/start.js:49` (and `src/common.js` flag list)
- **Severity:** Medium
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** The `--smart-snap*` flags are advertised on every storybook subcommand, but in `start`/URL mode `buildDir` is always `undefined`, so `applySmartSnap` immediately throws `SmartSnapBailError` and (with `failBuildOnFailure` false) silently falls back to a full snapshot set with only a single `log.info` line. Nothing in the flag descriptions warns that SmartSnap requires a built directory.
- **Suggestion:** Add an early `log.warn` when `smartSnap.enabled && !buildDir`, or note the build-dir requirement in the `smart-snap` flag description, or exclude the flags from the `start` command's flag list.

- **File:** `src/common.js:84-85`, `src/config.js:233-235`
- **Severity:** Low
- **Reviewer:** stack:percy-cli-code-review
- **Issue:** `statsFile` default `'enriched-stats.json'` is declared in three places (flag default, schema default, and re-defaulted inside `validateAndReadStats`). Not a precedence bug — all three agree and `env > flag > .percy.yml > default` is intact — but it is a future drift hazard. Same harmless duplication for the `default: false` booleans.
- **Suggestion:** Optional — keep the schema default as canonical and drop the flag-level `default` for `statsFile`, matching how other config-defaulted options rely on the schema.

- **File:** `src/utils.js:339-342` (story-store branch)
- **Severity:** Low
- **Reviewer:** stack:percy-cli-code-reviewer
- **Issue:** In the `__STORYBOOK_STORY_STORE__.raw()` branch, `docsEntries` are not concatenated, so `diagnostics.storiesTotal` reflects only raw stories in that path. Diagnostic-fidelity only (debug log), not a snapshot-behavior bug; pre-existing asymmetry preserved from master.
- **Suggestion:** Optional — note the asymmetry or include `docsEntries` consistently if diagnostics accuracy matters for debugging.

- **File:** PR (no `docs/`/changelog entry)
- **Severity:** Low
- **Reviewer:** stack:percy-cli-code-review
- **Issue:** A user-facing feature adding 7 flags + config keys ships with no docs or changelog note.
- **Suggestion:** Add a docs/product entry and surface the change in the PR description / changelog. Non-blocking.

---

**Verdict: PASS** — No Critical/High issues; security, CSP/browser-side hygiene, API precedence, and scope checks all pass, and the disputed `failBuildOnFailure` re-throw is confirmed intentional. The Medium test-coverage gap on the `snapshots.js` SmartSnap try/catch is strongly recommended before merge but is not a blocking correctness defect.

src/snapshots.js
```diff
@@ -333,6 +338,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
+    if (storybookConfig?.smartSnap?.enabled) {
+      try {
+        snapshots = yield applySmartSnap(percy, snapshots, storybookConfig.smartSnap, buildDir);
+      } catch (e) {
+        if (e instanceof SmartSnapBailError) {
+          log.info(e.message);
+        } else {
+          log.warn(`SmartSnap failed (${e.message}); running full snapshot set`);
+        }
+        if (storybookConfig.smartSnap.failBuildOnFailure) throw e;
```
pranavz28, line 350: **P1 — `failBuildOnFailure` fails the build on an intentional bail.**

This `throw e` runs for **both** branches above. A `SmartSnapBailError` is the engine's *designed* "run the full snapshot set" signal (matched `bailOnChanges` glob, no baseline, too many changes — `@percy/cli-command@1.32.0` `dist/smartsnap.js:632`). It's correctly logged at `info`, but then re-thrown when `failBuildOnFailure` is set — so strict mode **hard-fails on every normal bail**, which contradicts the flag's documented intent ("fail the build if SmartSnap encounters any *error*"). A bail is not an error; it already falls back to the full set.

Move the re-throw into the non-bail branch only:
```js
} else {
  log.warn(`SmartSnap failed (${e.message}); running full snapshot set`);
  if (storybookConfig.smartSnap.failBuildOnFailure) throw e;
}
```
_(Flagged independently by 3 reviewers — strong consensus.)_

pranavz28 (commented):
Automated review (multi-agent: correctness, security, architecture, quality/testing, simplicity). Merge-gating is resolved — verified `@percy/cli-command@1.32.0` is published and its entry point exports both `applySmartSnap` and `SmartSnapBailError`. Design is sound: off by default, fails safe, and `importPath` is correctly stripped before capture (does not leak into published snapshots).

Flagging 2 P1 and 5 P2 findings inline. The headline item is the `failBuildOnFailure` re-throw on a normal bail.

src/snapshots.js
```diff
@@ -333,6 +338,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
     // set storybook environment info
     percy.client.addEnvironmentInfo(environmentInfo);
 
+    if (storybookConfig?.smartSnap?.enabled) {
```
pranavz28, line 341: **P1 — No test coverage on this orchestration block.**

The added `utils.js` tests cover `importPath` extraction well, but this integration path (the highest-risk new logic) is untested: off-default skip, success → `snapshots` reassigned to the filtered set, bail → `info` + continue, generic error → `warn` + fallback, and `failBuildOnFailure` re-throw. A single test here would have caught the P1 bug a few lines down.

Also add a guard test that `importPath` is stripped in `processStory` and never reaches `percy.snapshot()` (the leak-prevention is currently unverified by any test).

src/snapshots.js
```diff
@@ -333,6 +338,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
+      try {
+        snapshots = yield applySmartSnap(percy, snapshots, storybookConfig.smartSnap, buildDir);
```
pranavz28, line 343: **P2 — Validate untrusted inputs at this trust boundary; verify the engine contract.**

This forwards values to the engine with no validation: `importPath` (derived from in-browser Storybook data — untrusted if the Storybook is attacker-influenced), `statsFile` (joined with `buildDir`), and `baseline` (a git ref). Recommend, at this layer: reject `importPath` that escapes `buildDir` after `path.resolve` (and absolute paths / `..`); constrain `statsFile` to a basename; validate `baseline` against a conservative ref charset. Note `untraced`/`bailOnChanges` globs also bypass the repo's existing `MAX_PATTERN_LENGTH=500` guard used by `include`/`exclude`.

Also **verify before merge**: that `applySmartSnap` returns the filtered snapshots *array* and does **not** mutate the input array in place before throwing — the fail-safe fallback here depends on `snapshots` staying intact on error.

src/snapshots.js
```diff
@@ -333,6 +338,19 @@ export async function* takeStorybookSnapshots(percy, callback, { baseUrl, flags
...
+      } catch (e) {
+        if (e instanceof SmartSnapBailError) {
+          log.info(e.message);
+        } else {
+          log.warn(`SmartSnap failed (${e.message}); running full snapshot set`);
```
pranavz28, line 348: **P2 — Fallback `warn` logs only `e.message`, swallowing the stack.**

When SmartSnap fails it silently runs a full build, so the user may never notice it broke. Logging only `e.message` discards the stack and makes CI diagnosis hard. Add `log.debug(e)` (or `e.stack`) alongside the user-facing warn — consistent with the `log.debug` precedent for diagnostics elsewhere in this function.

src/snapshots.js
```diff
@@ -70,12 +70,12 @@ function shouldSkipStory(name, options, config) {
 // Returns snapshot config options for a Storybook story merged with global Storybook
 // options. Validation error messages will be added to the provided validations set.
 function getSnapshotConfig(story, config, invalid) {
-  let { id, ...options } = PercyConfig.migrate(story, '/storybook');
+  let { id, importPath, ...options } = PercyConfig.migrate(story, '/storybook');
```
pranavz28, line 73: **P2 — `importPath` coupling spans 5 sites; fragile to story-shape changes.**

The field is threaded through: serialize injection (`utils.js:386`), extraction (`utils.js:303-330`), this strip-before-validate, the re-merge below (`:78`), and the destructure in `processStory` (`:261`). Three of these exist only to shuttle one field past `PercyConfig.validate` (which rejects unknown props via `unevaluatedProperties:false`). If the story shape changes, all five must move in lockstep. Consider registering `importPath` as a known-internal field in the `/storybook` schema, or carrying it on a side-channel keyed by story `id` — at minimum, add cross-referencing comments at the 5 sites.

src/utils.js
```diff
@@ -312,14 +312,67 @@ export function evalStorybookStorySnapshots({ waitFor }, { docCapture = false, a
...
+      stories.__smartsnapDiagnostics = {
+        source,
+        entriesPresent: !!entries,
+        entriesCount: entries ? Object.keys(entries).length : 0,
+        storiesTotal: stories.length,
+        storiesWithImportPath: withImportPath,
+        sampleEntry,
+        sampleStory
+      };
     }
```
pranavz28, line 373: **P2 — `__smartsnapDiagnostics` is always-on debug instrumentation.**

This ~30-line blob runs whenever SmartSnap is enabled (not gated behind the existing `smart-snap-trace` flag), allocates sample objects, and scans all stories — its only consumer is one `log.debug` in `snapshots.js:327`. It also serializes raw story internals (`parameterKeys`, `fileName`, sample ids) into a debug line. Recommend deleting it, or collapsing to a single `trace`-gated line:
```js
if (smartSnap && trace) log.debug(`SmartSnap: ${stories.filter(s => s.importPath).length}/${stories.length} stories have importPath`);
```
Separately, attaching `stories.__smartsnapDiagnostics` as a property on the array works only by ordering luck (it won't survive a `.map`/`.filter`/`.concat` inserted later) — prefer a plain local variable returned as `{ stories, diagnostics }`.

src/storybook.js
```diff
@@ -41,6 +41,7 @@ export const storybook = command('storybook', {
 
   yield* takeStorybookSnapshots(percy, () => server?.close(), {
     baseUrl: args.url ?? server?.address(),
+    buildDir: args.serve,
```
pranavz28, line 44: **P2 — `buildDir` is `undefined` in `--url`/`start` mode, silently no-opping SmartSnap.**

The positional arg resolves to *either* `url` *or* `serve` (line 13), so for `percy storybook http://host/` (and the `start` command, which passes no `buildDir`) the engine can't locate the stats file. Best case it bails to a full run with only an `info` breadcrumb; the user enabled `--smart-snap` and gets no benefit with no clear explanation. Recommend an upfront `log.warn` (or validation) when `smartSnap.enabled && !args.serve` stating SmartSnap requires a build directory, not a URL.

src/snapshots.js
RaghavsBrowserStack: that's intentional. The failure is so the customer doesn't get charged unfairly, regardless of the source of the error.

src/snapshots.js
RaghavsBrowserStack: talked to security about this, because this code is running customer side this is acceptable. https://browserstack.slack.com/archives/CR42BQ3AQ/p1779964662139329

src/snapshots.js
RaghavsBrowserStack: bruh what. its ok I'll debug it don't worry

src/snapshots.js
RaghavsBrowserStack: yeah no, this is not some 50k + line codebase with multiple merges per day, moving something to state for code clarity will just make future debugging more difficult

src/utils.js
RaghavsBrowserStack: this one time flow doesn't need this level of over engineering

src/storybook.js
RaghavsBrowserStack: this is not silent, its the first thing mentioned in the docs

src/snapshots.js
RaghavsBrowserStack: again, for a 1 time flow if this a dealbreaker I don't mind adding it, but is it really @pranavz28

PR #1319, opened 2026-06-14
security: serialize Percy builds with an in-flight guard (PER-8546)

Shivanshu-07:
## Summary
Resolves [PER-8546](https://browserstack.atlassian.net/browse/PER-8546) (**High**, CWE-362, deadline 2026-06-16) and the concurrent-build leg of chain [PER-8634](https://browserstack.atlassian.net/browse/PER-8634).

## Problem
`registerSnapshotHandlers` fired `runPercyBuild()` for every `RUN_SNAPSHOT` event **with no `await` and no concurrency guard** (`snapshots.cjs:168`). Rapid events spawned N simultaneous Percy builds that race on shared globals (`process.env.PERCY_TOKEN`) and the `.env` file — corrupting the token mid-build and burning N× the comparison/CI quota billed to the org.

## Fix
Added a module-level `buildRunning` guard:
- set it before `runPercyBuild`, clear it in a `finally`;
- a `RUN_SNAPSHOT` received while a build is in flight is rejected with `SNAPSHOT_ERROR` (*"A Percy build is already in progress…"*) instead of starting a second build.

Serializing builds also eliminates the concurrent `process.env.PERCY_TOKEN` mutation race.

## Verification
`yarn`'s jasmine suite (`registerSnapshotHandlers` + handler-registration specs) passes, including a new test that:
- fires two `RUN_SNAPSHOT` triggers without awaiting the first → asserts the second is rejected with *"already in progress"*;
- confirms a later trigger (after the first completes) is allowed through (`SNAPSHOT_STARTED`).

Closes PER-8546; mitigates the concurrent-build leg of PER-8634 (the IDOR/credential leg of that chain is tracked with PER-8544).

[PER-8546]: https://browserstack.atlassian.net/browse/PER-8546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PER-8634]: https://browserstack.atlassian.net/browse/PER-8634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/snapshots.cjs
+++ b/src/server/snapshots.cjs
@@ -164,9 +164,28 @@ async function runPercyBuild(channel, { baseUrl, include = [], exclude = [] }) {
 
 /* ─── Channel handlers ─────────────────────────────────────────────────── */
 
+// Serialize Percy builds. runPercyBuild mutates shared globals
+// (process.env.PERCY_TOKEN) and the .env file, so firing it concurrently for
+// rapid RUN_SNAPSHOT events races on that state, corrupts the token mid-build,
+// and spawns multiple billable builds (CWE-362). An in-flight guard rejects a
+// second trigger until the running build finishes.
+let buildRunning = false;
+
 function registerSnapshotHandlers(channel) {
-  channel.on(PERCY_EVENTS.RUN_SNAPSHOT, (data) => {
-    runPercyBuild(channel, data);
+  channel.on(PERCY_EVENTS.RUN_SNAPSHOT, async (data) => {
+    if (buildRunning) {
+      channel.emit(PERCY_EVENTS.SNAPSHOT_ERROR, {
+        message: 'A Percy build is already in progress. Please wait for it to finish before starting another.'
+      });
+      return;
+    }
+
+    buildRunning = true;
+    try {
+      await runPercyBuild(channel, data);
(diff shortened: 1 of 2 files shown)


Shivanshu-07:
## Impact analysis + automated review

**Impact — no existing functionality breaks.** The in-flight guard serializes `RUN_SNAPSHOT`/builds, which for the interactive Storybook devserver flow is single-user/serial anyway. The flag is set synchronously before the first `await` and cleared in `finally` (no deadlock), and a concurrent trigger is surfaced as `SNAPSHOT_ERROR` rather than silently dropped. The previous parallelism was the bug (token/`.env` corruption), so no legitimate throughput is lost.

**`/stack:pr-review`: PASS** — no Critical/High findings (concurrency fix verified correct). Minor follow-ups (non-blocking): drop the ticket id from the new test description; consider an `afterEach` reset of the module guard for test isolation.

pranavz28 (approved):

